PR #192, opened 2015-01-07
changed "has" to "had"

alex-sherwin:
When read as "has" it makes an ambiguous argument that JGroups currently "has" these deficiencies, which is the opposite point of what the statement is trying to make


rachmatowicz:
Alex

Thanks for the corrections to the text.

Generating one PR for changing a few characters at a time is expensive. A lot of manual work has to be done to process these PRs: reviewing, merging, closing. How about combining all your typographical changes for the docs into one large PR and submitting it instead?

alex-sherwin:
Sure, sorry I've just been making PR's as I'm reading through the whole set
of documentation.

I will close the individual PR's and make one PR

PR #200, opened 2015-01-22
JGRP-1876 Strange mergeview when the merge leader misses a coord MergeInfo

kammous:
When the merge leader misses a coord MergeInfo, mereView doesn't contain that coord.


belaban:
I think this fix makes things worse with large subgroups, see my comment on JGRP-1876.
If you want to submit a new PR, make sure you submit it from a branch of yours not from master.

PR #207, opened 2015-04-20
BZ1211227 Extracting version dynamically

slaskawi:
Hey @belaban !

Please take a look at proposed solution of https://bugzilla.redhat.com/show_bug.cgi?id=1211227.

During JDG development we use our private fork and it is very easy to forget about updating the version. With "automatic" approach we won't have to do it manually.

Could you please double check if I used the correct Classloader? I thought about using the one, which loaded the `Version` class (`manifestAsStream = Version.class.getResourceAsStream("/version.properties");`)

Thanks!

### DIFF
--- a/conf/jgroups-version.properties
+++ b/conf/jgroups-version.properties
@@ -1,0 +1,1 @@
+version=${project.version}
--- a/pom.xml
+++ b/pom.xml
@@ -110,6 +110,13 @@
             <resource>
                 <directory>conf</directory>
                 <includes>
+                    <include>jgroups-version.properties</include>
+                </includes>
+                <filtering>true</filtering>
+            </resource>
+            <resource>
+                <directory>conf</directory>
+                <includes>
                     <include>*.xml</include>
                     <include>jg-messages*.properties</include>
                 </includes>
--- a/src/org/jgroups/Version.java
+++ b/src/org/jgroups/Version.java
@@ -5,6 +5,12 @@ package org.jgroups;
 
 import org.jgroups.annotations.Immutable;
 
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Properties;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 /**
  * We're using the scheme described at http://www.jboss.com/index.html?module=bb&op=viewtopic&t=77231
  * for major, minor and micro version numbers. We have 5 bits for major and minor version numbers each and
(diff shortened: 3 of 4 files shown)


belaban:
Not sure I like this: it's based on maven and uses an additional file, whereas the JGroups build system uses ant/ivy. Also, what's the reason JDG needs to use a private fork ?
I'll take a more detailed look when I return from my PTO

dereed:
My comments:
- Don't use a generic name like "/version.properties"!
  Either put it in a package (org/jgroups/version.properties), or give it a unique name (jgroups-version.properties).

- Don't ignore exceptions.  If there's a problem setting the version, that's fatal.

- The regex is looking for ".Final".  What about non-final versions?

- Bonus points for setting the description to the full version string (for example 3.4.4.Final-redhat-5-bz-123456) instead of the stripped down version (3.4.4.Final), so that the full version is printed from "java -jar jgroups.jar" or Probe.

slaskawi:
Hey @belaban !

Thanks for looking into this change. This is not a top priority issue, so please enjoy your PTO and take a look at it when you get back :)

* Why a property file?

I think this is the most convenient way of dealing with versions. Alternatively we could filter the sources and treat one of the java files as a template but the biggest disadvantage is that it would be hard to run the tests from IDE. Maybe there is some other/better way of dealing with such things in Ant?

* Why do we use a private fork in JDG?

Sometimes we need to put in only some small bits during the minor release and we don't want to perform a full upgrade. It is very convenient to use our own fork in such circumstances.

slaskawi:
Hey @dereed !

Thanks for the review. Please take a look at the updated PR and the comments below:

* Don't use a generic name like "/version.properties"!
 * Done
* Don't ignore exceptions. If there's a problem setting the version, that's fatal.
 * Done
* The regex is looking for ".Final". What about non-final versions?
 * I altered the regexp to `((\\d+)\\.(\\d+)\\.(\\d+)\\..*)`, but I'm not 100% sure what are kind of patterns are valid and invalid. Could you please give me a hand with that? Maybe this pattern should be more specific?
* Full version string (for example 3.4.4.Final-redhat-5-bz-123456)
 * I'm not sure how to extract this kind of name. `3.4.4.Final` is the easy part - it might be taken from `${project.version}`. `redhat-5` should be appended by Brew during the build (pom manipulation plugin changes the version on the fly, so it should be ok). But I have no idea how to get the last part :) Could you please give me a hint?

Thanks!

belaban:
Currently, we need to change the version in 4 places:
* Version.java
* build.xml
* pom.xml
* ivy.xml

Your suggested solution only works when using maven. However, the default build system for JGroups is ant & ivy. But I'm only going to accept a solution that works for both ant&ivy and maven.

For example, I like your idea of creating a new file VERSION which defines the version:
```
jgroups.version=3.6.4.Final
```
VERSION would be the only file that needs to be changed when changing the JGroups version. It is shipped with the JGroups JAR.
In `ivy.xml` and `build.xml`, VERSION could be included to define the version, also in `pom.xml`.
The `org.jgroups.Version` class would then have to parse VERSION to generate its major/minor/patch numbers and the version description.

belaban:
OK, so I added VERSION and changed build.xml and ivy.xml/ivysettings.xml

belaban:
OK, so I integrated the major parts of your PR, but haven't touched pom.xml yet. Can you take a look ? @slaskawi

slaskawi:
Hey @belaban ! Sure, no problem. I'll take a look later this afternoon.

belaban:
OK @slaskawi. We need to replace
```
<version>3.6.3.Final</version>
```
in pom.xml with the version extracted from `conf/VERSION.properties`. I don't know how to do it.
Cheers,

slaskawi:
Hey @belaban !

I've done some research and it turns out that changing `project.version` on the fly is very hard. Maven does not support reloading this property (this is mentioned in [1]). There are a couple of ideas how to solve this problem:
* Use Pom manipulation extensions [2], but they need to be installed in `MAVEN_HOME/lib/ext`. It is worth to mention that they are already installed in Brew.
* Use Maven invoker plugin and trigger Maven 2 times. The first run would change the version and the second will execute the build. Unfortunately this will result in pretty complicated `pom.xml`.
* The last solution requires changing our approach a little bit. How about setting project version in `pom.xml` and parsing it during Ant build? This approach seems the least complicated and I would vote for it.

That do you think?

[1] http://mojo.codehaus.org/versions-maven-plugin/usage.html
[2] https://github.com/release-engineering/pom-manipulation-ext#version-manipulation

belaban:
Approach 1 is not feasible, 2 would be up to you: as I said I don't use mavenas the build system in JGroups, so for my taste pom.xml can be as complex as it needs to be as long as no additional dependencies or file are needed. Approach 3 is not what I want as the version should not be defined in pom.xml.
Other approaches:
4. Generate pom.xml via ant. This would be simple; we'd have a pom.xml.template, use ant to copy it to pom.xml and then replace <version> via the ant replace task. `ant clean` would remove pom.xml
5. Remove `pom.xml` altogether. Hmm, I likey... :-)

belaban:
OK, I picked door 4: `ant jar`, `ant clean` (called by the former) or `ant make-pom` removes pom.xml and generates a new pom.xml from conf/pom.template.xml, replacing the version.

slaskawi:
Sounds great! I'll try to implement a wrapper pom which will do exactly the same. This way the end users would not need to call `ant make-pom; mvn clean install` but just `mvn clean install`.

Thanks!

belaban:
OK, then I could remove make-pom and conf/pom.template.xml. Awaiting your PR...

PR #213, opened 2015-05-27
JGRP-1930 RELAY2: SiteUUID and regular UUID are bundled by TransferQu…

pruivo:
…eueBundler

* Make sure that the SingletonAddress makes the distinction between SiteUUID and UUID.

https://issues.jboss.org/browse/JGRP-1930

note: I had to change the SingletonAddress. I couldn't change the compareTo from SiteUUID and UUID to return always different because jgroups was unable to find the physical address.

### DIFF
--- a/src/org/jgroups/protocols/relay/SiteUUID.java
+++ b/src/org/jgroups/protocols/relay/SiteUUID.java
@@ -1,5 +1,6 @@
 package org.jgroups.protocols.relay;
 
+import org.jgroups.Address;
 import org.jgroups.util.AsciiString;
 import org.jgroups.util.ExtendedUUID;
 import org.jgroups.util.UUID;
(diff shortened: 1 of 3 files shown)


belaban:
Did you verify that the change in {{SiteUUID.compareTo()}} doesn't have any negative effect in {{RELAY2}}/{{Relayer}} ?

pruivo:
your tests were green for me. when I first tried to a more aggressive ```compareTo``` (i.e.e returning ```!= 0``` when the we compare different ```Address``` classes) I got some failures and some "no physical address found" errors.

so, I'm pretty sure that it is fine :)

belaban:
OK. Make sure to run the xsite testsuite in Infinispan as well with the changes, to see if this works.
Cheers,thx,

pruivo:
infinispan x-site test suite is OK.

belaban:
Great, thx.
I'm closing the JIRA issue now

PR #217, opened 2015-06-09
BZ1211227 Store version in a single file

slaskawi:
Hey @belaban !

I've been trying to extract JGroups version for Maven build but it turns out that it is very complicated. I couldn't achieve satisfying results even with Maven invoker plugin. 

So I come up with an alternative approach - a bash script, which corrects the version in `pom.xml` and then invokes Maven. We can use it the same way as maven command directly: `./build.sh clean install`.

What do you think?

Thanks
Sebastian

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,118 @@
+#!/bin/bash
+
+PROGNAME=`basename $0`
+DIRNAME=`dirname $0`
+SED="sed"
+ROOT="/"
+MVN="mvn"
+
+#  OS specific support (must be 'true' or 'false').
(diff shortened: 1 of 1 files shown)


belaban:
Hi Sebastian: no, I don't want people to have to run a build.sh / build.bat file. I also didn't like my own solution with the make-pom target in the ant build.
I therefore reverted pom.xml. To change the version, conf/VERSION.properties and pom.xml have to be changed. Not an optimal solution, I know, but maven doesn't seem to be flexible enough to set the version dynamically, from a file.

PR #222, opened 2015-06-18
Moved to 4.0.0-SNAPSHOT

slaskawi:
Hey @belaban !

Please take a look at proposed changes in versioning.

Release repositories don't accept SNAPSHOTs by default and vice versa - snapshot repositories don't accept releases. Working on SNAPSHOT on daily basis will protect us from releasing artifacts by a mistake. Also this change is needed by CI SNAPSHOT job.

Thanks
Sebastian

### DIFF
--- a/conf/VERSION.properties
+++ b/conf/VERSION.properties
@@ -1,4 +1,4 @@
 
 ## Defines the JGroups version
 ## This file should be the *only place* to change when changing the version number
-jgroups.version=3.6.4.Final
+jgroups.version=4.0.0-SNAPSHOT
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
     <artifactId>jgroups</artifactId>
     <packaging>bundle</packaging>
     <name>JGroups</name>
-    <version>3.6.4.Final</version>
+    <version>4.0.0-SNAPSHOT</version>
     <url>http://www.jgroups.org</url>
 
     <properties>


conf/VERSION.properties
```diff
@@ -1,4 +1,4 @@
 
 ## Defines the JGroups version
 ## This file should be the *only place* to change when changing the version number
-jgroups.version=3.6.4.Final
\ No newline at end of file
+jgroups.version=4.0.0-SNAPSHOT
```
rhusar: Absolutely not! This needs to follow JBoss versioning rules!

belaban:
Absolutely not *what* ?

rhusar:
Needs to follow the https://developer.jboss.org/wiki/JBossProjectVersioning

belaban:
@slaskawi @rhusar What's the URL for uploading of snapshots ? And do you have the relevant section in `settings.xml` ?

conf/VERSION.properties
slaskawi: Could you please elaborate more why, and what kind of fix are you suggesting here? Unfortunately in JBoss versioning rules I couldn't find any paragraph dedicated to snapshots...

Looking at some examples [1][2][3] the proposed convention (`4.0.0-SNAPSHOT`) looks absolutely fine.

[1] https://repository.jboss.org/nexus/service/local/repositories/snapshots/content/org/hibernate/hibernate-c3p0/3.6.10-SNAPSHOT/maven-metadata.xml
[2] https://repository.jboss.org/nexus/service/local/repositories/snapshots/content/org/infinispan/infinispan-bom/7.2.2-SNAPSHOT/maven-metadata.xml
[3] https://repository.jboss.org/nexus/service/local/repositories/snapshots/content/org/drools/drools-bom/6.3.0-SNAPSHOT/maven-metadata.xml

slaskawi:
@belaban Here it is - JBoss SNAPSHOT repository: https://repository.jboss.org/nexus/content/repositories/snapshots/

belaban:
@rhusar: the qualifier is optional

belaban:
Hmm, reading the doc it should actually be `4.0.0.SNAPSHOT`, *not* `4.0.0-SNAPSHOT` (dot rather than hyphen)...

belaban:
OK, changed version to `3.6.4-SNAPSHOT`.
@slaskawi : how do I release a snapshot ? `mvn deploy` ? Does maven see that the version is a snapshot and therefore releases the snapshot to jboss-snapshots-repository instead of jboss-releases-repository ?

rhusar:
> how do I release a snapshot ?

It is differentiated automatically but the deploy plugin. If your project version ends with SNAPSHOT and you execute mvn deploy, artifacts will be deployed to your snapshot repository in xml emelent ```<snapshotRepository/>```.

conf/VERSION.properties
rhusar: Look at WildFly for instance where conventions tend to be followed first :-)
https://repository.jboss.org/nexus/#nexus-search;gav~org.wildfly~wildfly-parent~~~

Since Bela doesn't often do Alphas/CR for minors/micros, the best name would probably be "3.6.4.Final-SNAPSHOT", for major 4 it could probably start with a alpha so e.g. "4.0.0.Alpha1-SNAPSHOT".

tristantarrant: We don't necessarily have to follow the wildfly rules.

rhusar:
> @rhusar: the qualifier is optional


Have you looked at the "Current Release Version Conventions (including qualifiers) - Post January 8, 2010" section? It doesn't really seem to be optional..


> Hmm, reading the doc it should actually be 4.0.0.SNAPSHOT, not 4.0.0-SNAPSHOT (dot rather than hyphen)...

Hm, not really, since -SNAPSHOT is really just a maven concept. I am not if it must be a hyphen, but that is the convention in maven world that can break some tools...

conf/VERSION.properties
rhusar: There are no wildfly rules. I am talking about jboss rules https://developer.jboss.org/wiki/JBossProjectVersioning and maven convetions, both of which would be nice to follow.

PR #223, opened 2015-06-19
Fixes FILE_PING discovery method, when filename contains '/' (or '\' …

pimenas:
…on windows).

When the addressToFilename() returns a String with '/' (directory separator) FILE_PING discovery method cannot work. I found this issue while trying to use FILE_PING with infinispan server.

I simply replace File.separatorChar() with the character '-' and this fixes the problem.

### DIFF
--- a/src/org/jgroups/protocols/FILE_PING.java
+++ b/src/org/jgroups/protocols/FILE_PING.java
@@ -149,7 +149,8 @@ public class FILE_PING extends Discovery {
 
     protected static String addressToFilename(Address mbr) {
         String logical_name=UUID.get(mbr);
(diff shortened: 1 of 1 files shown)


belaban:
Where would `addressToFileName()` use a `File.separatorChar` ? Method `addressAsString()` certainly doesn't do that. Do you refer to `logical_name` ?

belaban:
Method `addressAsFilename()` is used in a lot of places (e.g. in `S3_PING`); I want to make sure this doesn't break things in other protocols.

pimenas:
Yes it is ``logical_name``, in my case it's value was ``infinispan-node-2/clustered``.

belaban:
ok

PR #224, opened 2015-06-27
JGRP-1909 Use org.jgroups.azure for the protocol ID instead

rhusar:
Hi Bela,

The clean approach is that extra projects should actually use org.jgroups.PROJECT package name, so that a repository jgroups-PROJECT contains the relevant classes/files. So e.g. in repo jgroups-azure shoud reside all classes in org.jgroups.azure package.

Cheers,
Rado


rhusar:
Yes, I am aware (and discussed with Paul too). Although, in WildFly/EAP people will have to specify which module to load this protocol from anyway, so this way, it might be clearer for users that they do need to specify a module.

The configuraiton will look like:

     <org.jgroups.azure.protocols.AZURE_PING module="org.jgroups.azure">
         <property name="account_key">...</property>
         ...
         
etc.

belaban:
OK, no problem on my side anyway

PR #227, opened 2015-07-08
Make X509Token look in the classpath for the keystore.

bertrama:
This behavior is consistent with how the ENCRYPT protocol looks for a keystore.


bertrama:
The proposed change here was copied from ENCRYPT.java.   Re-factoring the two classes might be better, but I can't say I'm invested enough to do that at the moment.

PR #228, opened 2015-07-14
Fix MERGE3 @Property for min_interval and max_interval

rhusar:
Unless I am missing something, the property annotation should move to the field instead of being twice on the get/set methods.

Noticed this when looking at the schema which was generated invalid because the properties were occuring twice, e.g.:

```xsd
      <xs:element name="MERGE3">
        <xs:complexType>
          <xs:attribute name="min_interval" type="xs:string"/>
          <xs:attribute name="min_interval" type="xs:string">
            <xs:annotation>
              <xs:documentation>Minimum time in ms before sending an info message</xs:documentation>
            </xs:annotation>
          </xs:attribute>
...
```

This makes the schema valid again. T. Hanks.

### DIFF
--- a/src/org/jgroups/protocols/MERGE3.java
+++ b/src/org/jgroups/protocols/MERGE3.java
@@ -38,8 +38,11 @@ public class MERGE3 extends Protocol {
     
 
     /* -----------------------------------------    Properties     -------------------------------------------------- */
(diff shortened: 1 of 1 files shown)


belaban:
Thx Rado !

PR #234, opened 2015-09-04
BZ1247250: Jgroups: null: logical address cache didn't contain all physical address

rpelisse:
Hi Bela,

As discussed over IRC, as a fix for this issue, I've switched notification to trace - which is matching what you end up doing upstream (more/less). Let me know if you feel we should handle this issue differently.


rpelisse:
(for instance, maybe you want to keep the log on WARN for non shared transport - if it is the case, just say so, I'll update the PR)

rpelisse:
Close PR as I've reworked it to remove useless whitespaces change

PR #238, opened 2015-09-18
Added option to not wrap exceptions with InvocationTargetException

kedzie:
Wasn't sure what branch to push this to, I didn't see one with 3.6.7-SNAPSHOT.  So i'll let you guys figure that out.  

I encountered a problem with wrapped exceptions between JDK 6 (Android-18) & JDK 7.   ```RequestCorrelator``` wraps *thrown* exceptions with ```InvocationTargetException```.. which has had the same ```serialVersionUID``` since Java 1.1.  But on JDK 1.7 ```InvocationTargetException extends ReflectiveOperationException``` *(which has a different serialVersionUID)*. Exceptions thrown are serialized and I get an exception because the serialVersionUID doesn't match.

```
java.io.InvalidClassException: java.lang.ReflectiveOperationException; Incompatible class (SUID): java.lang.ReflectiveOperationException: static final long serialVersionUID =123456789L; but expected java.lang.ReflectiveOperationException: static final long serialVersionUID =2505831918131413020L;
at java.io.ObjectInputStream.verifyAndInit(ObjectInputStream.java:2343)
at java.io.ObjectInputStream.readNewClassDesc(ObjectInputStream.java:1643)
at java.io.ObjectInputStream.readClassDesc(ObjectInputStream.java:657)
at java.io.ObjectInputStream.readNewClassDesc(ObjectInputStream.java:1663)
at java.io.ObjectInputStream.readClassDesc(ObjectInputStream.java:657)
at java.io.ObjectInputStream.readNewObject(ObjectInputStream.java:1784)
at java.io.ObjectInputStream.readNonPrimitiveContent(ObjectInputStream.java:761)
at java.io.ObjectInputStream.readObject(ObjectInputStream.java:1985)
at java.io.ObjectInputStream.readObject(ObjectInputStream.java:1942)
at org.jgroups.util.Util.objectFromByteBuffer(Util.java:501)
```
```123456789L``` is the ```serialVersionUID``` of ```ReflectionOperationException``` in JDK 7

The default behavior is unchanged.  It is only changed if you set the system property to true.

Returning exceptions is not really an option in my application as there are already return values.  If you can suggest a better solution I'm all ears.

### DIFF
--- a/src/org/jgroups/Global.java
+++ b/src/org/jgroups/Global.java
@@ -105,7 +105,7 @@ public class Global {
     public static final int IPV4_SIZE=4;
     public static final int IPV6_SIZE=16;
 
-
+    public static final String DONT_WRAP_EXCEPTIONS = "jgroups.dont.wrap.exceptions";
 
 
     public static boolean getPropertyAsBoolean(String property, boolean defaultValue) {
--- a/src/org/jgroups/blocks/RequestCorrelator.java
+++ b/src/org/jgroups/blocks/RequestCorrelator.java
@@ -449,6 +449,7 @@ public class RequestCorrelator {
     protected void handleRequest(Message req, Header hdr) {
         Object        retval;
         boolean       threw_exception=false;
+        boolean       wrap_exception=System.getProperty(Global.DONT_WRAP_EXCEPTIONS, "false").equals("false");
 
         if(log.isTraceEnabled()) {
             log.trace(new StringBuilder("calling (").append((request_handler != null? request_handler.getClass().getName() : "null")).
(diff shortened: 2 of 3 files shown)


kedzie:
Any thoughts on this guy?  I would love for it to get merged in... @belaban

belaban:
Sorry, haven't been able to take a look yet, been teaching JGroups workshops in NYC and Calif. Will take a look soon

belaban:
Hmm, not so good... the test fails because I'd have to reset the system prop in a the `BeforeMethod` annotation...

belaban:
OK, added wrap_exceptions to MessageDispatcher and RequestCorrelators. Take a look

kedzie:
In the last test you actually set ```wrap_exceptions``` to true instead of false... probably want to fix this; I'm making a PR, it was easy fix.

PR #241, opened 2015-10-15
Fix regression with ForkChannel

johnou:
When the ForkChannel constructor is called with create_fork_if_absent as
true a FORK protocol should be created if it does not exist. After
fba4f23282c281873a09256b038f2251829596ba
org.jgroups.protocols.FORK#createForkStack sets the channel from stack
which is never set resulting in an NPE..

channel = new ForkChannel(baseChannel,
                            "hijack-stack",
                            "lead-hijacker",
                            true,
                            ProtocolStack.ABOVE,
                            FRAG2.class);

Caused by: java.lang.NullPointerException
	at org.jgroups.protocols.FORK.createForkStack(FORK.java:199)
	at org.jgroups.fork.ForkChannel.<init>(ForkChannel.java:79)

### DIFF
--- a/src/org/jgroups/fork/ForkChannel.java
+++ b/src/org/jgroups/fork/ForkChannel.java
@@ -271,6 +271,7 @@ public class ForkChannel extends JChannel implements ChannelListener {
             if(!create_fork_if_absent)
                 throw new IllegalArgumentException("FORK not found in main stack");
             stack.insertProtocol(fork=new FORK(), position, neighbor);
+            fork.setProtocolStack(stack);
         }
         return fork;
     }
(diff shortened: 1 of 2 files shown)


src/org/jgroups/fork/ForkChannel.java
```diff
@@ -271,6 +271,7 @@ protected static FORK getFORK(Channel ch, int position, Class<? extends Protocol
             if(!create_fork_if_absent)
                 throw new IllegalArgumentException("FORK not found in main stack");
             stack.insertProtocol(fork=new FORK(), position, neighbor);
+            fork.setProtocolStack(stack);
```
johnou: @belaban on second thought it might be safer to actually construct and set the protocol stack before inserting the protocol into the stack?

```java
fork=new FORK();
fork.setProtocolStack(stack);
stack.insertProtocol(fork, position, neighbor);
```

PR #243, opened 2015-10-20
Call callAfterCreationHook before init.

johnou:


belaban:
Done. I also added a few more calls to the creation hook, e.g. after dynamically inserting / replacing a protocol

PR #245, opened 2015-11-12
Enhanced GMS join retry loop to include discovery

onagano-rh:
This is just a proof of concept which works as expected in a multiple segments network.
The very first real coordinator will take more time to start thought.

https://issues.jboss.org/browse/JGRP-1977

### DIFF
--- a/src/org/jgroups/protocols/pbcast/ClientGmsImpl.java
+++ b/src/org/jgroups/protocols/pbcast/ClientGmsImpl.java
@@ -81,9 +81,10 @@ public class ClientGmsImpl extends GmsImpl {
 
             responses.waitFor(gms.join_timeout);
             responses.done();
+            join_attempts++;
             long diff=System.currentTimeMillis() - start;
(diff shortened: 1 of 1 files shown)


belaban:
I don't like this as the max time for the coord to start could be as high as `max_join_attempts * join_timeout`.
Discovery is an unreliable process, and if we don't get a result the first time, then we become a singleton member and merge later. This is the design.
Besides, I don't want to taint the design of GMS which doesn't know (and doesn't need to know) how discovery works.

belaban:
If we wanted to make an unreliable discovery process more reliable, we could perhaps do this in `PING`, e.g. by sending multiple discovery requests and spacing them out over the `join_timeout` period.
This is the way it was done, but this code was subsequently removed because I didn't like it: either code is completely reliable (retransmit until success or dest crashes), or unreliable, but not something in-between.
What you could do is create your own protocol (below `PING`) and do this logic in your protocol.

PR #246, opened 2015-11-19
JGRP-1968: JGroups missing JBoss logging internationalization

rpelisse:
Externalizing a bunch more messages - done by a script, so some little fiddling may be required afterward, but most generation issue have been catched.


belaban:
Applied, with changes

belaban:
Oh man, do I hate this crap ! I really don't believe in internationalization of error messages...
The automatic replacement leads to stupid things like this:
```
log.error(Util.getMessage("RequestOfType") + req.type + 
                                " sent an object of " + objectToRun + " which is invalid");
```
Because we're only internationalizing the first part, we could have an error message whose first part is French and the rest English...
Do you want to fix this manually ?
Using a resource bundle can lead to runtime exceptions if arguments are supplied incorrectly to resource error strings.

rpelisse:
Yes, I'll work on that tomorrow, and do an other PR to remove this sort of i18n messages.

rpelisse:
Hi Bela,

Would it not make more sense to turn this kind of message to accept parameter, like this one:

```java
JGRP000034: %s: failure sending message to %s: %s
```

This way will avoid the scenario where a message is "half translated" (and it will be more of an enhancement). I'm going to do that this afternoon and send a PR, but let me know if you don't to change the message like that....

belaban:
No, that's exactly what I want and what I was referring to in my earlier comment.

PR #247, opened 2015-11-19
JGRP-1968: JGroups missing JBoss logging internationalization - removed a couple of unused messages

rpelisse:
Found a couple of unused messages using the following scripts:

```bash
#!/bin/bash

readonly SRC=${SRC:-'./src'}
readonly PROPS_FILE=${PROPS_FILE:-'conf/jg-messages.properties'}

cut -f1 -d\   "${PROPS_FILE}" | \
while
  read messageId
do
  grep -q -e "${messageId}" -r "${SRC}"
  if [ "${?}" -ne 0 ]; then
    echo "${messageId}"
  fi
done
```


belaban:
done

PR #249, opened 2015-11-23
Fixed typo when checking views in GMS_MergeTest._testSimpleMerge()

bjc-world:
Hi Bela, minor fix to a GMS unit test


belaban:
Thanks !

PR #250, opened 2015-11-24
JGRP-1968: JGroups missing JBoss logging internationalization

rpelisse:
Fixed some messages to use the printf format, and also remove a duplicate message.

@belaban Note that I've managed to remove a duplicate message (Header Type %s is not know") BUT while FC uses the message, it does not have the local_addr filled to respect the formatting of the message. 

We could have a duplicate message just for this class, but instead I've opted to use the msg.getSrc() value (more information is always better than more). Let me know what you think if you don't feel it is such a good idea.


belaban:
There's a number of errors, e.g. 
`FailedToDeliverOOBMessage = JGRP000218: failed to deliver OOB message %s` is passed the excepion, but there's no second `%s`. I'm changing this to
`FailedToDeliverOOBMessage = JGRP000218: failed to deliver OOB message %s: %s`

belaban:
I had to change quite a number of message formats, as most of the changes were missing the exception at the end !
I'm a bit worried that this will introduce bugs on log error/warn messages. Can you double-check that all error message formats match the actual error messages?

PR #251, opened 2015-11-30
SimplifiedTransferQueueBundler

rvansa:
* does not use collections to sort destinations
* uses ArrayBlockingQueue as the main queue

The new bundler works for me, though, I haven't found any tests that I should adapt and I have encountered errors when running `mvn test`:
` [testng] An error occurred while instantiating class org.jgroups.tests.byteman.ServerTest: org.jgroups.util.Util.close(Ljava/net/ServerSocket;)V`
So I haven't ran these.

### DIFF
--- a/src/org/jgroups/protocols/TP.java
+++ b/src/org/jgroups/protocols/TP.java
@@ -1269,7 +1269,12 @@ public abstract class TP extends Protocol implements DiagnosticsHandler.ProbeHan
             else if(bundler_type.startsWith("transfer-queue") || bundler_type.startsWith("new")) {
                 if(bundler_type.startsWith("new"))
                     log.warn(Util.getMessage("OldBundlerType"), bundler_type, "transfer-queue");
-                bundler=new TransferQueueBundler(bundler_capacity);
+                // TODO: remove system option
+                if (bundler_type.endsWith("simplified")) {
+                    bundler = new SimplifiedTransferQueueBundler(bundler_capacity);
+                } else {
+                    bundler = new TransferQueueBundler(bundler_capacity);
+                }
             }
             else if(bundler_type.startsWith("sender-sends")) {
                 bundler=new SenderSendsBundler();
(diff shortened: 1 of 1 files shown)


belaban:
Taking a look. I like the idea of getting rid of the hashmap in favor of a simple array, but...
* First benchmarks indicate this is slightly slower than TransferQueueBundler. I used 4 {{UPerf}} instances on the same box
* The hard limit of {{512}} now possibly triggers a bundled send() sooner than {{max_bundle_size}}; if we send a lot of very small messages, this will be suboptimal. Let's see if we can find a solution to this...
* When the array size (512) has been reached, the current sender sends all bundled messages. This is similar to the {{sender-sends}} bundler type, but now you have a problem: {{sendBundledMessages()}} needs to be protected, as both the bundler thread and a caller's thread can call this method concurrently!

Which perf test did you use to get the double-digit (iirc?) perf increase?

rvansa:
I have used JMH benchmark that simulates 5 or 10 producers and single (S)TQB consumer. The code is a bit different; I had to refactor it for JMH and testing and then back, though the logic should be the same.
Here is the benchmark and adapted TQB:
https://github.com/rvansa/benchmarks/blob/master/ispntest/perftest/src/main/java/org/jboss/perf/hibernate/BundlerBenchmark.java
https://github.com/rvansa/benchmarks/tree/master/ispntest/perftest/src/main/java/org/jgroups/protocols
I have to try out UPerf as well to see the difference.

The hard limit is a guestimate - that's obviously something that could be configurable, but there's already too many knobs in JGroups configuration so I haven't added another one :) My concern was that if that array was too long, multiple iterations through it would lower the performance. And I am not optimizing for ultra-short messages as it's rather a corner case for Infinispan. Though, further suggestions are welcome.

Wrt `sendBundledMessages`: I don't think you're right, the sender threads call only `send(Message)` which enqueue the message (and blocks without space in the queue), while the array is operated by the bundler thread only.

belaban:
Ah, ok, I thought `addMessage()` was called by the adder, but obviously that not the case. I'll take another look at the hard-coded array length tomorrow (was out of office today).

belaban:
Applied your changes with modifications (mainly style). 2 things to revisit:
* The hard coded 512 sized buffer
* Moving bundlers out of TP. I want to be able to (e.g.) define the fully qualified classname of a bundler to be used.

rvansa:
@belaban Just FYI, I've ran two simple ISPN benchmarks (4 nodes, 50 threads executing exclusively puts or gets, TCP stack) and I got these data:

Before:
265,388 gets/s, 753 us MRT
96,564 puts/s, 2.06 ms MRT

After:
268,457 gets/s, 744 us MRT
97,647 puts/s, 2.04 ms MRT

It seems that there's also some CPU usage reduction, but generally this is a marginal difference in performance. I'll have to run JFRs yet to see how memory consumption has changed; though, the effect won't be linear given the speedup (and especially as ISPN generates way lot of garbage).

belaban:
Is this IspnPerfTest, JMH or Radargun?
For now, `SimplifiedTransferQueue` is in, but I'm not advertizing it yet (docs), so we can perf-test it, so no hurries.

rvansa:
RadarGun. Sure, no need to recommend it one day after merge :)

PR #253, opened 2015-12-10
JGRP-1992 GOOGLE_PING discovery fails if port 443 isn't used

alanfx:
Change GOOGLE_PING to use port 443 when creating a connection and no
port is specified in the configuration.

### DIFF
--- a/src/org/jgroups/protocols/GOOGLE_PING.java
+++ b/src/org/jgroups/protocols/GOOGLE_PING.java
@@ -17,8 +17,9 @@ public class GOOGLE_PING extends S3_PING {
     }
 
     protected AWSAuthConnection createConnection() {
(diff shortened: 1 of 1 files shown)


belaban:
Thanks for the fix, Alan. Did you test this with a real GCE account, or did you use some simulator?

alanfx:
I tested this with a real GCE account, Bela.

belaban:
OK. I read that there is a Google Cloud server that can be run off of the local file system to check a client's correctness.
Might be worth taking into consideration for future unit testing...

alanfx:
That would be interesting. Do you have a link with more information?

belaban:
https://cloud.google.com/datastore/docs/tools/devserver

alanfx:
Thanks a lot, Bela

PR #254, opened 2015-12-12
Improved documentation

MatiasSM:
As discussed in http://sourceforge.net/p/javagroups/mailman/message/34689441/


belaban:
Thanks Matias,

PR #255, opened 2015-12-17
Fixed log format error in PERF class.

ginaluca:
Added missing parameter in a couple of log statements.


belaban:
Thx Gianluca !

PR #259, opened 2016-01-14
squid:S00115 - Constant names should comply with a naming convention

m-ezzat:
This pull request is focused on resolving occurrences of Sonar rule squid:S00115 - Constant names should comply with a naming convention
https://dev.eclipse.org/sonar/coding_rules#q=squid:S00115
M-Ezzat

### DIFF
--- a/src/org/jgroups/Global.java
+++ b/src/org/jgroups/Global.java
@@ -21,8 +21,8 @@ public class Global {
 
     public static final short SCOPE_ID=ClassConfigurator.getProtocolId(SCOPE.class);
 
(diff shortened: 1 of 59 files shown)


belaban:
Your change breaks backwards compatibility, e.g. `Version.version` to `Version.VERSION` breaks code that uses this. Also, changing public enums also breaks client code.
I'm not following the Java coding conventions, only where it makes sense. I don't think uppercase constants increase readability when everyone uses an IDE nowadays with coloring of constants.
And - yes - I use `my_var` underscore variable, so don't even think about changing them to `myVar`... :-)

PR #261, opened 2016-01-15
squid:S2259 - Null pointers should not be dereferenced

m-ezzat:
This pull request is focused on resolving occurrences of Sonar rule squid:S2259 - Null pointers should not be dereferenced
https://dev.eclipse.org/sonar/coding_rules#q=squid:S2259
M-Ezzat


belaban:
I made modifications for the false positives, e.g. the method below can never throw an NPE because str is never accessed:

````
 public static int sizeUTF(String str) {
        int len=str != null? str.length() : 0;
        for(int i = 0; i < len; i++) {
           int  c=str.charAt(i);
         ...
    }

PR #264, opened 2016-01-15
Remove Base64 from S3_PING and add latest from iHarder.net to org.jgroups.util

bjc-world:
Hi Bela

I've removed the Base64 static inner class from S3_PING and added the latest from iHarder.net (v2.3.9 according to comments in the code) into the util package and checked that the tests pass.

Cheers
Ben


belaban:
Hi Ben,
your PR is out-of-whack (has a lot of conflicts); can you pull&rebase and re-submit?

PR #265, opened 2016-01-16
Remove Base64 from S3_PING and add latest from iHarder.net to org.jgroups.util

bjc-world:
Hi Bela

I'm afraid I'm new to git, so my apologies if this doesn't work - I've tried rebasing my fork and pulled the latest changes, but I'm not sure if I've done it correctly (says my branch is 4 commits ahead, but 3 of those are merge PRs from belaban/master).

Ben


belaban:
OK, I added the latest Base64 from iharder.net to org.jgroups.util and removed inner class Base64 from S3_PING.

PR #268, opened 2016-01-19
squid:S2162 - "equals" methods should be symmetric and work for subcl…

m-ezzat:
This pull request is focused on resolving occurrences of Sonar rule squid:S2162 - equals methods should be symmetric and work for subclasses
https://dev.eclipse.org/sonar/coding_rules#q=squid:S2162
M-Ezzat

### DIFF
--- a/src/org/jgroups/View.java
+++ b/src/org/jgroups/View.java
@@ -144,7 +144,7 @@ public class View implements Comparable<View>, Streamable, Iterable<Address> {
     }
 
     public boolean equals(Object obj) {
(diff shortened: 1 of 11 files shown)


belaban:
In some cases (UUIDs / SiteUUIDs and Views / MergeView), I want to run an superclass.equal(subclass). With your changes, this would break.

m-ezzat:
and what do you think if we replace 'this' by 'super', can it solve those cases ?

belaban:
I don't know. To be honest, we ran into a subtle issue with equals() on UUID and SiteUUID/SiteMaster, so I don't want to change that code as it works perfectly.

PR #270, opened 2016-01-26
Fix maven build with Java 8

jmarkos:


belaban:
Applied on branch JGRP-2007, will merge shortly. Thanks!

PR #277, opened 2016-02-01
improve performance for Request.cancel

schaebo:
This pull-request fixes serious performance problems we encountered with 3.6.7-final when cancelling a lot of requests at once. 

The Problem can be tracked down to the RequestCorrelator, more exactly `this.requests.values().remove(req)`.
If the request Map has a lot of entries, removing by value gets really slow. 

The basic idea is to keep the request ID for each Request as part of the Object, so it can be used later to remove the value from the Map by key instead of value.

### DIFF
--- a/src/org/jgroups/blocks/Request.java
+++ b/src/org/jgroups/blocks/Request.java
@@ -36,9 +36,9 @@ public abstract class Request implements NotifyingFuture, org.jgroups.util.Condi
     protected volatile boolean        done;
     protected boolean                 block_for_results=true;
     protected volatile FutureListener listener;
+    private long req_id;
 
 
-    
     public Request(Message request, RequestCorrelator corr, RequestOptions options) {
         this.request_msg=request;
         this.corr=corr;
@@ -181,5 +181,11 @@ public abstract class Request implements NotifyingFuture, org.jgroups.util.Condi
         return cond.waitFor(this, timeout, TimeUnit.MILLISECONDS);
     }
 
+    public void setRequestId(final long req_id) {
+        this.req_id = req_id;
+    }
 
+    public long getRequestId() {
+        return req_id;
+    }
 }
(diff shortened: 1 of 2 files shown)


belaban:
Ouch, another field for every request... :-) Let me see whether/how I can optimize this, e.g. I'm thinking of removing `request_msg` and passing it via `execute()`. That would then not increase the size of request instances, and perhaps I can apply this to other fields as well.

belaban:
Also, this field is unneeded for asynchronous requests, which are not added to the request table in `RequestCorrelator`...

schaebo:
Thank you for the fast reaction.

Ok, I can see why the additional field poses a problem, especially if it is unused for some requests.

The problem manifested itself when calling cancel() and if i unterstand your proposal correctly it will improve "normal" request->response, but not the "cancel - case" ?

belaban:
Yes, I thought `cancel()` wasn't called often and hence didn't optimize it. I'll see if I can bypass the `RequestCorrelator` for async requests altogether, then I wouldn't mind adding a request-id field to a request, as we'd be dealing only with _sync_ requests.

belaban:
OK, so I removed `request_msg` from `Request` and pass it to `execute()`. Commit is [1]

[1] 924f97f7851d1c95673664012f3a69edbc9bcae5

belaban:
So it turns out `Request` actually _had_ a `req_id` field, but I removed it on Jan 8 2016 (commit 55d1272d) as part of an optimization.
My design made the tradeoff between a fast `cancel()` operation and a compact in-memory layout in favor of the latter.
What's the use case you have? Why do you have a lot of cancellations?

belaban:
Oops, I'm an idiot! {{RequestCorrelator.done()}} is also called whenever a request is complete, not just on cancellation! Ignore my comment above, investigating further...

schaebo:
I'm pretty sure you're not an idiot ;-)

Our use case is a timeout for the response, because if the response takes longer than X it might not be valid/useful anymore, so we cancel the original request and then send a new request. 
In short a timeout => retry scenario, in which we need to be aware of the timeout and be able react to it (also changing the request)

belaban:
OK, it hurts but I added `req_id` back to `Request`... :-)
JIRA: https://issues.jboss.org/browse/JGRP-2013

belaban:
I'm going to commit after running the test suites on master and 3.6.8 (master running right now), please verify that this is the solution you wanted. Note that I changed quite some code, especially on master...
I don't think I need to `RequestCorrelator.done(Request r)` any longer, as all requests added to `RequestCorrelator.requests` are also going to have a request ID (async requests are not added anyway).

PR #280, opened 2016-02-05
squid:S2446, squid:S2325 - "notifyAll" should be used, private method…

m-ezzat:
This pull request is focused on resolving occurrences of Sonar rule
squid:S2446, squid:S2325 - "notifyAll" should be used, private methods that don't access instance data should be static

You can find more information about the issue here: https://dev.eclipse.org/sonar/coding_rules#q=squid:S2446
https://dev.eclipse.org/sonar/coding_rules#q=squid:S2325

M-Ezzat


belaban:
Applied with changes: using `notifyAll()` instead of `notify()` is a performance drag, as we want to wake up exactly *one* thread (which ones doesn't matter).

PR #284, opened 2016-02-10
JGRP-2015 Fixes around new DONT_BUNDLE+OOB processing:

rhusar:
Maybe all this is intentional and this makes no sense, but maybe:

DONT_BUNDLE+OOB 
* messages are not removed from batch when execution fails
* rejections are not counted to num_rejected_msgs
* spurious logging per message overfills the log
* treat failures other than RejectedExecutionException as errors (as batch processing does)


belaban:
Also backported to the 3.6 branch

rhusar:
Thanks!

PR #286, opened 2016-02-12
JGRP-2014 FILE_PING destination file name can include File.separator

alanfx:
characters

Replace File.separatorChar with '/' to fix this issue on Windows


belaban:
Applied. I actually changed this slightly to replace _both_ backslash and slash with a hyphen

belaban:
Backported to 3.6 branch

PR #288, opened 2016-02-16
JGRP-2014 FILE_PING destination file name can include File.separator

alanfx:
characters

Filter out all illegal filename characters in
FILE_PING.addressToFilename()
Use platform specific line endings when writing to a file

### DIFF
--- a/src/org/jgroups/protocols/FILE_PING.java
+++ b/src/org/jgroups/protocols/FILE_PING.java
@@ -141,8 +141,7 @@ public class FILE_PING extends Discovery {
     protected static String addressToFilename(Address mbr) {
         String logical_name=UUID.get(mbr);
         return (addressAsString(mbr) + (logical_name != null? "." + logical_name + SUFFIX : SUFFIX))
-          .replace(File.separatorChar, '-')
-          .replace('/', '-');
+          .replaceAll("[\0<>:\"/\\|?*]", "-");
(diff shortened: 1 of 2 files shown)


belaban:
Indeed a nice solution, thanks Alan!

src/org/jgroups/protocols/FILE_PING.java
```diff
@@ -141,8 +141,7 @@ protected boolean addDiscoveryResponseToCaches(Address mbr, String logical_name,
     protected static String addressToFilename(Address mbr) {
         String logical_name=UUID.get(mbr);
         return (addressAsString(mbr) + (logical_name != null? "." + logical_name + SUFFIX : SUFFIX))
-          .replace(File.separatorChar, '-')
-          .replace('/', '-');
+          .replaceAll("[\0<>:\"/\\|?*]", "-");
```
johnou: how often is this invoked? should we consider extracting the regex into its own static field so it is only compiled once?

belaban: Good idea, done (master and 3.6 branch).
Wow, that was quick, someone *is* paying attention here! :-)

alanfx: Good suggestion, @johnou! Lots of eyeballs and all...

PR #292, opened 2016-02-23
JGRP-1993 Compare the encoded keys instead of the SecretKey implementation

tristantarrant:
- fixes failures with IBM JDK

(master)


belaban:
Thanks Tristan, much appreciated!

PR #297, opened 2016-03-03
JGRP-2022 XSD schemas do not properly validate

rvansa:
All schemas now use namespace urn:org.jgroups:SOMETHING:VERSION where
SOMETHING is one of jgroups, fork and relay, and VERSION is aligned with
major.minor JGroups version and published on
http://www.jgroups.org/schema/SOMETHING-VERSION.xsd (it is not necessary
to publish new schema if it has not changed since the previous version,
though; older schema can be still used even in newer version).
Note that once published (in x.x.0 release), the schema should not be further
changed until next minor version.

https://issues.jboss.org/browse/JGRP-2022

### DIFF
--- a/conf/EncryptKeyStore.xml
+++ b/conf/EncryptKeyStore.xml
@@ -3,8 +3,8 @@
 <!-- input file: jGroupsEncryptedServerConfig.xml -->
  
 <config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-        xmlns="urn:org:jgroups"
-        xsi:schemaLocation="urn:org:jgroups http://www.jgroups.org/schema/jgroups.xsd">
+        xmlns="urn:org.jgroups:jgroups:3.6"
+        xsi:schemaLocation="urn:org.jgroups:jgroups:3.6 http://www.jgroups.org/schema/jgroups-3.6.xsd">
     <UDP/>
     <PING/>
     <MERGE3/>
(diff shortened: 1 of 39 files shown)


belaban:
Hmm, problems:
* The generated schema doesn't validate
* fork.xsd doesn't exist, it is still fork-stacks.xsd
* fork-stacks.xsd: cannot resolve type `jgroups:ConfigType`

rvansa:
OK, I somehow missed the position of xs:any in generated schema. And I can rename fork-stacks.xsd  to fork.xsd, but there's a problem that makes current fork stacks configuration non-validable. You'd prefer to paste the stack as it is into the fork-stacks.xml, into the config element, but even though you declare config element of the type jgroups:ConfigType, you'd still need to prefix every protocol with jgroups:, which is quite inconvenient.
I am working on a way to change this, but not without breaking compatibility.

rvansa:
Also, I've found that embedding the schema into main jgroups config does not work as you'd like...

rvansa:
So, the schemas above works with `xmllint --nonet --noout --schema schema.xsd file.xml`, but in order to test this you need to change `<xs:import schemaLocation="...">` to refer to the local file, not the one already published on web.
Please check `external.xml` which shows how the embedded configurations look. Not really nice, but I did not want to rework schema generator.
Also note that I haven't touched the parsers, so parsing is probably broken.

belaban:
sigh... this is getting too complicated. Let's just leave this then, XSD schema has always worked for me... :-)

belaban:
Or we discuss it in Konstanz...

PR #303, opened 2016-04-13
JGRP-2046 Order of XSD protocol elements and protocol attributes is arbitrary

rhusar:
https://issues.jboss.org/browse/JGRP-2046

With this also backported to 3.6.x its pretty cool, here is a report I can now make of configuration differences between 3.6.8.Final and 4.0.0-SNAPSHOT:

https://gist.github.com/rhusar/67abf87afdd556700a424b35772ad67a


belaban:
Thanks Rado!

PR #309, opened 2016-06-01
JGRP-2072 UFC and MFC headers get mixed up

danberindei:
https://issues.jboss.org/browse/JGRP-2072

Create separate FcHeader instances for UFC and MFC.

More of a workaround than a proper fix: the message includes the protocol id, not just the header id, so there should be no way to confuse the headers.

### DIFF
--- a/src/org/jgroups/protocols/FlowControl.java
+++ b/src/org/jgroups/protocols/FlowControl.java
@@ -28,11 +28,6 @@ import java.util.concurrent.TimeUnit;
  */
 @MBean(description="Simple flow control protocol based on a credit system")
 public abstract class FlowControl extends Protocol {
-
-    protected final static FcHeader REPLENISH_HDR=new FcHeader(FcHeader.REPLENISH);
-    protected final static FcHeader CREDIT_REQUEST_HDR=new FcHeader(FcHeader.CREDIT_REQUEST);  
-
-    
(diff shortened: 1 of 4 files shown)


belaban:
No, this is a proper fix, see my comment on JGRP-2072

belaban:
Applied and merged onto 3.6 branch, backported to master as well

danberindei:
Ah, I added a commit to change the default {{max_block_time}}, but I was too slow :)

src/org/jgroups/protocols/FlowControl.java
```diff
@@ -46,7 +41,7 @@
      * a REPLENISHMENT request to the members from which we expect credits. A value <= 0 means to wait forever.
      */
     @Property(description="Max time (in ms) to block")
-    protected long max_block_time=5000;
+    protected long max_block_time=500;
```
johnou: was this intentional or for testing?

johnou: intentional I guess.. 886273eb504ba94c956856178e351020802975e6 :)

danberindei: It was intentional, but I only added the commit after Bela had already integrated the pull request.

belaban:
Wow John, you're quick! :-)
Yes, Dan wanted this to be shorter, so senders don't get blocked for 5s in the worst case...

PR #311, opened 2016-07-15
Ignore lock requests from members not known in our current view

bramklg:
This fixes a (rare) bug we had on cluster splits/merges where two nodes used the same lock but where not (yet) present in the current view. This caused deadlocks. Ignoring requests from members not in the current view fixed the problem. It only happens when using UDP as transport protocol.

I tried to create an isolated testcase for this but failed to do so. I can't properly determine if there are side effects with this patch but as far as I can tell there shouldnt be any.


belaban:
The log statement was incorrect, fixed this and applied manually.

PR #312, opened 2016-07-19
Remove unaccessible staging repository

jmarkos:
The maven build is broken (mvn install), because https://repository.jboss.org/nexus/service/local/staging/deploy/maven2 cannot be accessed (I don't know why not, haven't found any information about that repo online). JGroups builds fine without it, though.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -84,21 +84,6 @@
         </repository>
 
         <repository>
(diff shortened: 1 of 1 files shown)


belaban:
Thanks Jakub!

PR #315, opened 2016-08-04
Update JDBC_PING.java to use try-with-resource to ensure that ResultS…

mway-dmueller:
…ets are closed automatically after they've been used.


belaban:
Thanks (and sorry for the delay, was on PTO)!

PR #318, opened 2016-08-15
fixes JDBC_PING.readAll() executing ResultSet.deleteRow() on readonly ResultSet

t-beckmann:
After updating JGroups the JDBC_PING constantly printed the failed removing row; please delete it manually error message. The cause was my program not exiting cleanly before updating, leaving rows in the cluster table. After the update, the PingData of the previous version failed deserialization and JGroups attempted deleting the offending row. Since the PreparedStatement was created readonly using the default parameters of Connection.prepareStatement(), the call resulted in an error causing the message printout.

The fix creates the statement as updatable so that the row can be deleted.

### DIFF
--- a/src/org/jgroups/protocols/JDBC_PING.java
+++ b/src/org/jgroups/protocols/JDBC_PING.java
@@ -262,8 +262,22 @@ public class JDBC_PING extends Discovery {
         }
     }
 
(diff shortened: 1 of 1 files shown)


belaban:
Thanks Thomas!

PR #325, opened 2016-09-30
SLF4J support

kgusarov:
Hi!

I've added very simple SLF4J (http://www.slf4j.org/) support as a default logger. This change seems to be very simple and straightforward, however for me it was a must-have feature. I am open to discussions.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -114,6 +114,11 @@
             <version>(2,)</version>
             <optional>true</optional>
         </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>[1.7,)</version>
+        </dependency>
 
         <!-- Test dependencies -->
         <dependency>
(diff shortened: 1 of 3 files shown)


belaban:
JIRA: https://issues.jboss.org/browse/JGRP-2105

I don't like making slf4j the default, so I introduced system property "jgroups.log_class" which is the fully qualified classname of the Log impl.

Is there no way to change the logging level at runtime? This is frequently used when we debug a system and change the log level with probe, and then back when done

kgusarov:
Hi! 

I've made slf4j default since it is facade that delegates all the dirty work to the underlying logging system e.g. log4j, logback etc. That is also the reason it won't allow to change the level in the runtime.

Configuration property sounds fine also. It is possible to introduce slf4j after it's done IMHO.

belaban:
OK, this is on master now. I also updated the documentation. Check it out. You need to set -Djgroups.log_class though

PR #336, opened 2017-01-09
JGRP-2139 SRV entries parsing for DNS_PING

slaskawi:
https://issues.jboss.org/browse/JGRP-2139

This PR introduces DNS discovery based on [`SRV` entries](https://www.ietf.org/rfc/rfc2782.txt). 

In Kubernetes/OpenShift this allows fine grained discovery (coarse grained is based on Headless Services) upon exposed ports. In order to get this working, one needs to expose a port with a **name** as shows in [this example](https://github.com/slaskawi/jgroups-dns-ping-example/blob/master/service.yaml#L19). An obtained `SRV` entry is used for a standard `A` discovery.


slaskawi:
This PR can be tested using these instructions: https://github.com/slaskawi/jgroups-dns-ping-example#testing-srv-entries

PR #338, opened 2017-01-13
Jgrp 1388/dns ping documentation

slaskawi:
https://issues.jboss.org/browse/JGRP-1388

Additional documentation for `DNS_PING`.

Hey @belaban! 

This is the last PR for `DNS_PING`. I created a PR (and not pushed directly) because I wanted you to take a look if I didn't miss any spot where we need additional documentation.

Cheers!

### DIFF
--- a/doc/manual/protocols.adoc
+++ b/doc/manual/protocols.adoc
@@ -747,6 +747,51 @@ This will use a bucket "jgroups-bucket" or create one if it doesn't exist, then 
 ${GOOGLE_PING}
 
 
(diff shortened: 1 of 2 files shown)


belaban:
Hi @slaskawi Looks good. Wouldit make sense though to add 1-2 sentences on when to use DNS_PING and when KUBE_PING, and perhaps a short explanation of the diffs?

slaskawi:
Good point! I will add it. Thanks!

PR #339, opened 2017-01-13
Build: run make-schema ant target in headless java process

rhusar:


belaban:
Small change / great effect! rado.incrementBeers()...

rhusar:
I had to, it was pissing me off that the new processes was stealing focus when building jgroups :-)

belaban:
Exactly what I was thinking, too!

PR #340, opened 2017-01-24
[master] JGRP-2153 Encrypt protocols transform empty buffer into null buffer when encrypt_entire_message is disabled

pferraro:
https://issues.jboss.org/browse/JGRP-2153

### DIFF
--- a/src/org/jgroups/protocols/Encrypt.java
+++ b/src/org/jgroups/protocols/Encrypt.java
@@ -347,8 +347,7 @@ public abstract class Encrypt extends Protocol {
 
         // copy neeeded because same message (object) may be retransmitted -> prevent double encryption
         Message msgEncrypted=msg.copy(false).putHeader(this.id, hdr);
-        if(msg.getLength() > 0)
-            msgEncrypted.setBuffer(code(msg.getRawBuffer(),msg.getOffset(),msg.getLength(),false));
+        msgEncrypted.setBuffer(code(msg.getRawBuffer(),msg.getOffset(),msg.getLength(),false));
         down_prot.down(msgEncrypted);
     }
 


belaban:
Paul, I don't think this will work if the buffer is null!
How about just setting the buffer to "" if it is it's size is 0 or it is null?

belaban:
Not sure I like this fix: we'll encrypt null or empty messages and also have to decrypt them. This is additional work, that would not be needed if your Marshaller was able to handle null buffers...

pferraro:
When is the buffer ever null?  Remember, a null RPC response does not translate to a null buffer.

pferraro:
I disagree, this is a matter of symmetry.  If a sender encrypts and sends an empty buffer, the receiver should receive and decrypt the message as an empty buffer - not a null buffer.

belaban:
If the buffer is never null, why do we get NPEs on the de-serialization?

belaban:
I agree on the symmetry argument

belaban:
Paul, you're not the only one sending messages down a channel. So if an app happens to use a different Marshaller, or an app uses the JChannel directly to send a message with a null buffer, we'll get an exception trying to encrypt the message!

pferraro:
Surely it is always up to the Marshaller to be able to unmarshal any buffer that it marshalled.  I don't think we should expect a Marshaller to be able to unmarshal a value that it never marshalled!

Anyway - I'll modify this commit to additionally handle null buffers.

pferraro:
Actually, I like the idea of bypassing encrypt/decrypt for empty buffers.  I'll modify.

pferraro:
"If the buffer is never null, why do we get NPEs on the de-serialization?"
Because the encrypt protocol converts what was an empty buffer into a null buffer.  This is the crux of the problem.

belaban:
I think the PR will cause problems returning a null encrypted message: when sign_msg is true, this will cause an NPE.
OK, so I think I'll do the following:
* When a buffer is null -> don't encrypt a message and pass it down. On the receiver side, a null buffer is not decrypted and passed up as well (assuming encrypt_entire_message is false in both cases)
* When the buffer is empty -> don't encrypt and pass down, same on the receiver side
* Otherwise enrypt when sending and decrypt when receiving

pferraro:
OK - feel free to close this.

pferraro:
Hang on a sec - I don't see how this PR would cause any new NPEs.  It looks like this protocol was not implemented with null buffers in mind.  All of the encrypt_entire_message and sign_msg logic assumes a non-null buffer.
Bela, are you sure that a null buffer is even a valid use case in JGroups?  Would we even be talking about this if the *_ENCRYPT protocols weren't erroneously creating a null buffer from an empty one?

belaban:
Yes, null buffers are a valid use case; e.g. some app sending a message with a application-defined header and a null buffer.
You're right, creating a null message from an empty (non-null) buffer wreaks havoc, and my suggestion in my previous comment will fix this.

pferraro:
OK - sounds like we're on the same page.  I'll close this and eagerly await your fix. :)

belaban:
OK. Just don't be too eager, I have other stuff on my plate... :-)

PR #341, opened 2017-01-24
[3.6] JGRP-2153 Encrypt protocols transform empty buffer into null buffer when encrypt_entire_message is disabled

pferraro:
https://issues.jboss.org/browse/JGRP-2153


pferraro:
Closing, per https://github.com/belaban/JGroups/pull/340

PR #342, opened 2017-02-07
JGRP-2063 JDBC_PING: allow supplying dataSource object directly inste…

rhusar:
…ad of String-based properties

https://issues.jboss.org/browse/JGRP-2063

Reviving this as @pferraro is refactoring the way we build the protocol stack which is required for further integration of JGroups and Elytron in WildFly so there will be use for this.


rhusar:
@pferraro could you ack this? Thanks

PR #343, opened 2017-02-09
Message delivered up the stack and throwing an exception: a warning with

ppalaga:
the exception name rather than the entire stack trace is logged
(https://issues.jboss.org/browse/JGRP-1958)

This is a manual backport of
https://github.com/belaban/JGroups/commit/03d7c93145f6a9314c2380f5ad66bdc8ec4209f0
for EAP 6.4 https://bugzilla.redhat.com/show_bug.cgi?id=1399195

@belaban  could you please review if I backported correctly?

### DIFF
--- a/src/org/jgroups/protocols/UNICAST2.java
+++ b/src/org/jgroups/protocols/UNICAST2.java
@@ -712,7 +712,7 @@ public class UNICAST2 extends Protocol implements AgeOutCache.Handler<Address> {
         xmit_reqs_sent.addAndGet(missing.size());
     }
 
-    
+
     /**
      * Called by AgeOutCache, to removed expired connections
      * @param key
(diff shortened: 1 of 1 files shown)


src/org/jgroups/protocols/UNICAST2.java
```diff
@@ -766,7 +766,7 @@ protected boolean handleDataReceived(Address sender, long seqno, short conn_id,
                 up_prot.up(evt);
             }
             catch(Throwable t) {
-                log.error("couldn't deliver OOB message " + msg, t);
+                log.error("couldn't deliver OOB message " + msg + ": " + t);
```
johnou: instead you might consider the following pattern..

```java
if (log.isDebugEnabled()) {
   log.error("couldn't deliver OOB message " + msg, t);
} else {
   log.error("couldn't deliver OOB message " + msg + ": " + t);
}
```

belaban:
@johnou I like that, and will keep it in mind for changes in master. But this is only a backport to an obscure and old branch nobody uses... :-)

PR #349, opened 2017-05-22
Cleanup AUTH design doc; upgrade to markdown

rhusar:


belaban:
Actually, I prefer asciidoc over markdown, so if we decide to convert more txt files, let's pick asciidoc instead.

rhusar:
Noted. Since these are rendered on github, any of the supported styling by GitHub is fine, but surely we can respect your preference in sync with the formatting used for docs. 
I wasn't primarily after upgrading to proper formatting on GitHub but I wanted to fix a few issues in the doc and thought, why not pretty print in the process :-)

belaban:
Agreed

PR #350, opened 2017-06-14
Remove elements of cluster when coordinator stops

douglasryanadams:
Currently, the JDBC_PING fails to clean up the database table like the FILE_PING cleans up its cluster file when the coordinator shuts down, this creates confusion when bringing the cluster back up sometimes as the new host will get stuck trying to delete a coordinator that no long exists in the database table.

This fix attempts to align the JDBC_PING functionality more closely with FILE_PING.

### DIFF
--- a/src/org/jgroups/protocols/JDBC_PING.java
+++ b/src/org/jgroups/protocols/JDBC_PING.java
@@ -110,6 +110,11 @@ public class JDBC_PING extends FILE_PING {
         attemptSchemaInitialization();
     }
 
+    @Override
+    public void stop() {
+        super.stop();
+        if (is_coord) removeAll(cluster_name);
+    }
 
     protected void write(List<PingData> list, String clustername) {
         for(PingData data: list)
--- a/src/org/jgroups/protocols/JDBC_PING.java
+++ b/src/org/jgroups/protocols/JDBC_PING.java
@@ -113,7 +113,7 @@ public class JDBC_PING extends FILE_PING {
     @Override
     public void stop() {
         super.stop();
-        if (is_coord) removeAll(cluster_name);
+        removeAll(cluster_name);
     }
 
     protected void write(List<PingData> list, String clustername) {


belaban:
Hi Douglas,

this change doesn't take into account attributes `remove_all_data_on_view_change` and `remove_old_coords_on_view_change`. I suggest we change your PR to something like this:
```` java
    if(remove_all_data_on_view_change)
        removeAll(cluster_name);
     else if(remove_old_coords_on_view_change) {
         Address old_coord=old_view != null? old_view.getCreator() : null;
         if(old_coord != null)
             remove(cluster_name, old_coord);
    }
````

What do you think?

**Comment by belaban** [comment]
If you agree, can you commit another PR or amend the existing one?

**Comment by douglasryanadams** [comment]
Hello, I'd be happy to make that update if that's all that's required but I am also curious about the intent of FILE_PING and JDBC_PING. Are they supposed to continuously represent the current state of the clusters or is it intended to be a record of all members who have ever connected (with unique new unique entries on restarts)?

It looks like both FILE_PING and JDBC_PING will 'collect' new records for every member restart. Even though members might remove themselves from the list on shutdown, the coordinator appears to re-add them in short order. I suspect that there's some underlying conflict between how Discovery.java (or perhaps something that uses it) is written to support the more traditional mutlicast/UDP discovery mechanisms and how FILE_PING.java and JDBC_PING.java actually manage their records.

If the intention is to keep the file / database table in sync with the state of the cluster I can try to track that down and create a more robust pull request this weekend. If things are generally working as expected, I'll take your suggestion and add that.

Thanks! I appreciate the feedback :)

**Comment by belaban** [comment]
The discovery protocol don't usually keep old members in their store, but some protocols (e.g. the cloud-based ones) try to minimize access to the cloud store so old members are not purged when they leave until the coord writes the new view, or a manual purge is started.

So, yes, the ideal scenario is that where the discovery store has all current members. TCPPING would typically not fulfill that if not all members are listed.

A member restart means that the member's UUID changes, so there will be a new entry with a new UUID but the same IP address (and possibly the same port, too).

Take a look at [1] for details.

[1] http://www.jgroups.org/manual4/index.html#DiscoveryProtocols

**Comment by douglasryanadams** [comment]
Hello,

I've spent some time playing around, looking at things, and making sense of your recommended revision.

The lines you proposed are already implemented in the parent class (FILE_PING) in the method `handleView` which will be called by `down` as I imagine was intended. By that mechanism, I believe `remove_all_data_on_view_change` and `remove_old_coords_on_view_change` are already  being taken into account in the context of JDBC_PING.

With that in mind, I reviewed up my initial PR and have concluded that this is actually more in line with FILE_PING: 

```java
    @Override
    public void stop() {
        super.stop();
        removeAll(cluster_name);
    }
```

In FILE_PING itself, these are the relevant calls:

```java
    public void stop() {
        super.stop();
        stopInfoWriter();
        remove(cluster_name, local_addr);
    }
    protected void remove(String clustername, Address addr) {
        if(clustername == null || addr == null)
            return;

        File dir=new File(root_dir, clustername);
        if(!dir.exists())
            return;

        log.debug("remove %s", clustername);

        String filename=addressToFilename(addr);
        File file=new File(dir, filename);
        deleteFile(file);
    }
```

As you can see, it really just dumps the whole file named after the cluster, I intend to do the equivalent with JDBC_PING in this PR.

Does that make sense?


I played around with an entirely different workflow but I think I was lacking enough of the big-picture to make sense of everything, there's a lot going on as I'm sure you're aware. (You can see that experiment here: https://github.com/douglasryanadams/JGroups/blob/jdbc_ping2/src/org/jgroups/protocols/JDBC_PING2.java ) I got stuck trying to retrieve the physical address from the stack.

belaban:
The patch you submitted looks fine; removing the entire table is probably ok as the new coordinator writes it again anyway. However, clearing and re-populating the table is not very efficient, plus we could lose data if the new coord doesn't have information about all members...

Do you think that (re-)use of `remove_all_data_on_view_change` might make sense here? If true -> removeAll(), else no-op?

douglasryanadams:
I see what you mean, that's why initially I was thinking you may only need to remove the cluster when the coordinator stops. My hesitation to include the `remove_all_data_on_view_change` comes from the knowledge that it appears to have a different intent that's already implemented in the parent class FILE_PING. Since omitting the `removeAll()` logic creates an inconsistent experience when restarting a whole cluster, setting `remove_all_data_on_view_change` to *false* would probably only cause the bug to show up again.

If we want to limit the times it interacts with the database we could return to my original patch, where it only makes the change when the coordinator is shutting down (though I suppose that approach might be more fragile if a coordinator crashes).

belaban:
`FILE_PING` only removes the member which shuts down (gracefully), so if you remove all rows in the table on coord shutdown, that would change the semantics.

Not that I disagree with this *for JDBC_PING only*, but I'm concerned that not making this configurable means everybody has to pay for this.

Consider `{A,B,C,D}`, with `A` leaving, `B` leaving, `C` leaving etc. Every time, this would cause a clear-table (old coord), fill-table sequence (new coord)...

douglasryanadams:
Is that the intention? The behavior I observe with FILE_PING is that every member sticks around in the file for that cluster until the coordinator shuts down which then deletes the whole file. If there's still another coordinator online it will then pick up with a new file and will (curiously) include the old coordinator in that file but as a member.

It appears to do this because every time `protected void write(List<PingData> list, String clustername)` is called `List<PingData>` appears to include all of the members the cluster's ever seen.

Perhaps there's something else going on? I had suspected their was in-memory cache that the Discovery mechanism was keeping and then publishing to the File (or Database depending) but when I looked for that cache mechanism I couldn't find it. Do you know what controls the `List<PingData>` content during `write`?

belaban:
I meant that when a coord leaves (gracefully), it'll clear the table and the next coord will fill the table again.
Yes, old information about members is kept. The reason is potential merging later. Note that information in the logical addres cache (TP) is only purged when either the max cache size has been exceeded, or removed entries have exceeded a max lifetime.

belaban:
Before this thread gets too long: do you have a unit test or reproduceable test that shows what IYO is wrong?

douglasryanadams:
Thanks for taking the time to respond! A unit test wouldn't cover this scenario -- each of the "units" appears to function as intended, it's just that when used together they don't appear to produce the expected result. I suspect resolving the root of the issue may require a more comprehensive refactor if the load from all of the "extra" writes is concerning to you (it may be totally acceptable).

Some time this week I'll write up a comprehensive bug that details the scenario I'm describing both with FILE_PING and JDBC_PING so that you or someone on your team can track it.

jebl01:
>The behavior I observe with FILE_PING is that every member sticks around in the file for that cluster until the coordinator shuts down which then deletes the whole file. If there's still another coordinator online it will then pick up with a new file and will (curiously) include the old coordinator in that file but as a member.

Interesting... I have exactly the same problem (even worse since the new coordinator include **all** old entries)

Minutes ago, just before I found this discussion, I reported my issue here:
https://issues.jboss.org/projects/JGRP/issues/JGRP-2232?filter=allopenissues

@douglasryanadams any new findings since June? Did you complete your bug report?

douglasryanadams:
Hey @jebl01 - unfortunately, I don't have any new findings.

PR #351, opened 2017-06-22
[JGRP-2198] Reduce log level for RequestCorrelator "channel is not

ppalaga:
https://issues.jboss.org/browse/JGRP-2198 

@belaban could you please review? Esp. I am not sure if a similar change is required in UNICAST3.

### DIFF
--- a/src/org/jgroups/protocols/UNICAST2.java
+++ b/src/org/jgroups/protocols/UNICAST2.java
@@ -812,7 +812,7 @@ public class UNICAST2 extends Protocol implements AgeOutCache.Handler<Address> {
                 up_prot.up(evt);
             }
             catch(Throwable t) {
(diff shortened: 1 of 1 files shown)


belaban:
No issues with merging this as `UNICAST2` is deprecated anyway; `UNICAST3` is the default point-to-point retransmission protocol these days.

ppalaga:
@belaban thanks for merging. The question was whether `UNICAST3` suffers from the same issue and whether this fix should be ported also to `UNICAST3`. I suppose users are supposed to migrate from `UNICAST2` to `UNICAST3` over time. If so, we should ensure that all issues solved in the older `UNICAST2` will also be solved in the newer `UNICAST3` so that the migrating users do not face issues again, they reported and got fixed in an earlier version.

belaban:
Right. If you submit another PR against UNICAST3, I'll take a look.

PR #359, opened 2017-09-08
JGRP-2034 Use the JDK-specific SASL realm property so that it works w…

tristantarrant:
…ith IBM

https://issues.jboss.org/browse/JGRP-2034


belaban:
Backported to 3.6 branch

PR #366, opened 2017-10-18
Port protocol configuration archetypes from WildFly

pferraro:


belaban:
Sorry, but I don't like this! Providing an interface for just 1 implementation (JDBC, multicast) strikes me as over-engineering. If we have multiple impls, we can think of extracting common functionality into a _superclass_, like for example with `FlowControl` or `Discovery`.
What's the rationale for this?

pferraro:
We discussed the rationale for this during my presentation in Milan.  You seemed to be ok with it at the time, but I'll elaborate here in case you were sleeping... :)

The rationale, in a nutshell, is to eliminate the need for WildFly to extend JGroups protocols and thus deal with custom protocols (with distinct IDs).  There are several reasons why I needed to subclass a protocol:

1. The protocol does not expose a mechanism to configure itself with an externally provided capability without having to override a method, e.g. ASYM_ENCRYPT, JDBC_PING (although this was only a problem in JGroups 3.6).
2. One or more protocols share a common configuration aspect, but use different method names, or expect different types.  e.g. MPING.setMcastAddr(...) vs MPING.setMulticastAddress(...), TCPPING.setInitialHosts(List<PhysicalAddress>) vs TCPGOSSIP.setInitialHosts(List<InetSocketAddress>), etc.  Thus there is no single class/interface from which integrators can easily configure these protocols.
3. New versions of JGroups protocols typically duplicate configuration methods, but do not share a common superclass/interface.  Introducing configuration interfaces, even for a single object, future-proofs integrators from supporting new protocol versions.

While subclassing is often appropriate for common functionality (although delegation is a far more flexible solution to the same problem), it is not appropriate (or even feasible) for "aspects".  For example, both MPING and UDP support multicast configuration.  UDP is a transport while MPING is a discovery protocol.  UDP has more in common with TCP (i.e. they are both transport) while MPING has more in common with TCPPING (they are both discovery protocols) than they have in common with each other.  Refactoring JGroups so that MPING and UDP can extend a common superclass is absurd, just so they can share a configuration property.

The only case for subclassing in this PR that _could_ be made is TCPPING and TCPGOSSIP.  Both protocols extend Discovery - and thus we might create a common InitialHostsDiscovery subclass, however, these protocols would need to agree on a type of the initial host (i.e. PhysicalAddress vs InetSocketAddress) in order to of any be practical utility.  Additionally, these two initial hosts are not functionally equivalent.  One is a list of initial members while the other is a list of gossip router locations which manage the membership.  What these protocols _do_ have in common is that they are configured using a static list of externally addressed resources.  However, this is just a configuration aspect - there isn't any common functionality, other than the configuration properties themselves.

Of the interfaces in this pull request, only 1 has just a single implementation: JDBC_PING.  Rado added a public setDataSource(...) to JDBC_PING for 4.0, so [1] from above is no longer a concern.  If you can assure me that there will never be a JDBC_PING2, I can live without the need for a JDBCConfiguration interface.  However, looking forward, I can see a potential use case for a JDBC-based log for JGroups-RAFT, or maybe even a durable retransmission table, etc.  Additionally, JDBCConfiguration might be enhanced to add setUserName(...), setPassword(...) methods where the credentials can be supplied by, say, an elytron credential store.

pferraro:
@belaban OK - I've updated the PR based on our discussion.

"Read my lips, no new interfaces!"
                             -- Bela Ban, 2017

belaban:
OK, applied, committed and pushed

PR #372, opened 2017-12-12
Narrow set of thrown exceptions in Streamable and Marshaller method signatures.

pferraro:
This removes the need for consumers of these methods to consistently handle a non-specific java.lang.Exception - particularly useful when adapting to other marshalling frameworks.
Also, eliminates unnecessary use of reflection when reading objects of a known type from a stream.

### DIFF
--- a/src/org/jgroups/AnycastAddress.java
+++ b/src/org/jgroups/AnycastAddress.java
@@ -124,7 +124,7 @@ public class AnycastAddress implements Address, Constructable<AnycastAddress> {
 
     @Override
     public void readFrom(DataInput in) throws Exception {
-        destinations = (Collection<Address>) Util.readAddresses(in, ArrayList.class);
+        destinations = Util.readAddresses(in, ArrayList::new);
     }
 
 }
--- a/src/org/jgroups/AnycastAddress.java
+++ b/src/org/jgroups/AnycastAddress.java
@@ -65,6 +65,7 @@ public class AnycastAddress implements Address, Constructable<AnycastAddress> {
         }
     }
 
+    @Override
     public int serializedSize() {
         if (destinations == null) {
             return Global.INT_SIZE;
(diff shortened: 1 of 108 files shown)


belaban:
I like your first pull request. -1 on the second as I don't see what it brings to the table (show an example where this makes sense), plus it breaks existing code, ie. classes which implement Streamable.

belaban:
I cherry-picked c8f4f98, but I don't want the second commit for the reasons outlined above.

pferraro:
The intent of the 2nd commit is to align the method signature with every other API that needs to adapt to yours.  e.g.
https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsAddress.java#L61
https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTopologyAwareAddress.java#L123
https://github.com/wildfly/wildfly/blob/master/clustering/server/src/main/java/org/wildfly/clustering/server/group/AddressableNodeExternalizer.java#L44

Truthfully, it would go a long way if we could just change the method signature of the various Util.read*/write* methods - and leave the Streamable API intact.   Does that sound reasonable?  If so, I'll submit a separate PR.

belaban:
I already cherry-picked the changes to the various Util.read/write() methods, this is in master.
The other changes can only be done in 5.0, as they break existing code. If you want to create a JIRA to keep track of this, I can this myself once I've made JGRP-2218 master.

pferraro:
Sounds good.  I've opened https://issues.jboss.org/browse/JGRP-2242

belaban:
Perfect!

pferraro:
Just give me a heads up when you branch off 4.0.x from master.

belaban:
will do so

PR #379, opened 2018-02-03
Refactoring complex methods of classes CoordGsmImpl and ExecutionServiceDemo

santiago-a-vidal:
Related to no particular issue. This is a behavior-preserving refactoring.

Summary of this pull request:

- We are evaluating a research prototype called Bandago that assists in the refactoring of complex methods. Bandago is an Eclipse plugin that automatically identifies and refactors a type of code smell called Brain Method. A Brain Method centralizes the intelligence of a class and manifests itself as a long and complex method that is difficult to understand and maintain We have applied Bandago to your project, and we would like to receive feedback.

- Bandago is very conservative and you should not observe many source code changes (only in the affected class).

- The source code (after the refactoring) should behave equivalently to the original one.

- As a sanity check, we have run tests before and after Bandago performed the refactoring(s) on the project. All tests passed.

- In this case Bandago refactored the method ExecutionServiceDemo.loop() extracting fragments of its code into the new methods processLine, processSpecialCases, and changeSize. Also, the method CoordGmsImpl. handleMembershipChange was refactored extracting fragments of its code into the new methods createNewDigest and sendCurrentDigestAndMemberships

Thanks in advance for your help in this evaluation!

### DIFF
--- a/src/org/jgroups/demos/ExecutionServiceDemo.java
+++ b/src/org/jgroups/demos/ExecutionServiceDemo.java
@@ -219,104 +219,116 @@ public class ExecutionServiceDemo {
             if(line.startsWith("quit") || line.startsWith("exit"))
                 break;
 
-            if(line.startsWith("submit")) {
-                int randomNumbers = Integer.parseInt(line.substring("submit".length()).trim());
-                // Parse numbers and break into parts
-                byte[] numbers = new byte[randomNumbers];
-                
-                for (int i = 0; i < randomNumbers; ++i) {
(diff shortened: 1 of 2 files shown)


belaban:
I'm afraid I don't see how this improves the code... 
`CoordGmsImpl` is a central class and I don't want to apply changes that I haven't studied carefully, as they might introduce bug, but I don't have the time for this. I would accept changes to non-important classes, such as ExecutionServiceDemo, if you submitted them individually.

santiago-a-vidal:
Thanks a lot for your feedback.

PR #381, opened 2018-03-16
Test for MessageDispatcher's synchronous cast to actually block for r…

Sanne:
…eplies (failing)

This is a failing test; I believe it should behave differently? Would appreciate a review before I reach to JIRA.


Sanne:
@belaban I pushed another micro commit to amend the comments as they were now conflicting with reality :)

PR #385, opened 2018-05-16
Extended DNS_PING documentation

slaskawi:
Hey!

I extended a few bits on `DNS_PING` docs. Please have a look and tell me what do you think about it.

Thanks,
Seb

//cc @pferraro @rhusar @luck3y


luck3y:
@slaskawi Thanks for this, every bit of extra info helps :)

PR #386, opened 2018-05-17
INJECT_VIEW logging fix: extraneous new lines, duplicate protocol name

rhusar:
Though, to be honest, I am failing to see the usefulness of this protocol. While you can use it for splitting, this cannot be used for merging (which I tried to use this for). If you try to install back the previous merge after split, you are just going to end having to wait for merge to trigger anyway since this is just racy.

2018-05-17 17:01:23,024 INFO  [org.jgroups.protocols.INJECT_VIEW] (default task-1) [INJECT_VIEW] Received request to inject view node-1=node-1/node-2
2018-05-17 17:01:23,028 WARN  [org.jgroups.protocols.pbcast.NAKACK2] (thread-4,null,null) JGRP000011: node-2: dropped message 21 from non-member node-1 (view=[node-2|2] (1) [node-2])

### DIFF
--- a/src/org/jgroups/protocols/INJECT_VIEW.java
+++ b/src/org/jgroups/protocols/INJECT_VIEW.java
@@ -34,13 +34,13 @@ public class INJECT_VIEW extends Protocol {
     @ManagedOperation(description="Inject a view (example of view string format: A=A/B/C;B=B/C;C=C)")
     public synchronized void injectView(String newView) {
         try {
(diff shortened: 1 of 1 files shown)


rhusar:
cc @valdar

belaban:
Right; `INJECT_VIEW` is only used to trigger a split cluster to fall into subviews, so that `MERGE3` can merge them back together. AFAIC, it was never meant to be used to merge subclusters back into one...

PR #388, opened 2018-06-13
Apply JGRP-2234 in 3.6

andrefelixbr:
I've applied JGRP-2234 fix on 3.6.16-SNAPSHOT.

### DIFF
--- a/src/org/jgroups/Message.java
+++ b/src/org/jgroups/Message.java
@@ -550,7 +550,8 @@ public class Message implements Streamable {
     }
 
 
-    public Header getHeader(short id) {
+
+    public <T extends Header> T getHeader(short id) {
         if(id <= 0)
             throw new IllegalArgumentException("An ID of " + id + " is invalid. Add the protocol which calls " +
                                                  "getHeader() to jg-protocol-ids.xml");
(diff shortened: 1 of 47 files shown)


belaban:
Is that the smallest possible PR you could come up with? :-)

andrefelixbr:
The number the files changed is big, but many of them are related to (ProtocolStack.Position.ABOVE and BELOW).  It was int and now is an enum **Position**. Total usage: 59

   I changed files here JGroups/tests/byteman/org/jgroups/tests/byteman also because tests didn't pass.

- Byteman error: The value for dir attribute in annotation @BMScript is wrong. We have to add "conf" in the begin as it's the correct path.
      **from**:
       @BMScript(dir="scripts/ASYM_ENCRYPT_BlockTest", value="testASYM_ENCRYPT_NotBlockingJoin")
      **to**:
       @BMScript(dir="conf/scripts/ASYM_ENCRYPT_BlockTest", value="testASYM_ENCRYPT_NotBlockingJoin")

belaban:
Sorry, but this PR is way too big and makes things fail: it introduces a dependency on junit, NameCache (instead of UUID) etc.
I want a minimal change to this very stable branch, only backporting the changes that are really needed, e.g. Protocol.Position does not need to be backported!
Please submit a new minimal PR only related to JGRP-2234. Doing a cherry-pick on JGRP-2234 doesn't work, as it introduces too many transitive changes!

PR #389, opened 2018-06-21
Test to demonstrate MERGE3 speed on tcp stack

rhusar:
Hi Bela,

here is a manual test for the issue I brought up at last meeting in Milan with MERGE3 when using TCP stack. The test configures Infinispan-like stack, creates partition by installing a view and waits for partition to merge back. This yields results like this one (runs exactly the same setup and only switches TCP/UDP):

```
-------------------------------------------------------------------
Results for TCP
Min merge time=48034
Max merge time=101050
Avg merge time=86958.1
===================================================================
Results for UDP
Min merge time=48019
Max merge time=48104
Avg merge time=48056.0
===================================================================
```

The problem being that merges with fairly low intervals 10-30 seconds can take actually up to 2 minutes for merge. Let me know what you think.

### DIFF
--- a/src/org/jgroups/protocols/VERIFY_SUSPECT.java
+++ b/src/org/jgroups/protocols/VERIFY_SUSPECT.java
@@ -76,6 +76,17 @@ public class VERIFY_SUSPECT extends Protocol implements Runnable {
     public VERIFY_SUSPECT() {       
     }
 
+    /* ------------------------------------------ Builder-like methods  ------------------------------------------ */
+
+    public VERIFY_SUSPECT setTimeout(long timeout) {
+        this.timeout = timeout;
+        return this;
+    }
+
+    public long getTimeout() {
+        return timeout;
+    }
+
     public Object down(Event evt) {
         switch(evt.getType()) {
             case Event.SET_LOCAL_ADDRESS:
--- a/tests/other/org/jgroups/tests/InfinispanStackMerge3Test.java
+++ b/tests/other/org/jgroups/tests/InfinispanStackMerge3Test.java
@@ -1,0 +1,128 @@
+package org.jgroups.tests;
+
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Supplier;
+import java.util.stream.Stream;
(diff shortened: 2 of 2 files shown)


belaban:
Hmm, interesting: UDP and SHARED_LOOPBACK have more or less the same numbers: it takes roughly `check_interval` ms for a merge to complete. TCP, OTOH, takes up to 100% more time. Investigating...

rhusar:
@belaban Thanks for looking into it, let me know if I can help.

belaban:
The reason TCP takes longer than UDP is that TCP needs multiple rounds to merge all 10 members, whereas UDP always only needs one round. Need to investigate why this is the case...

belaban:
Found out what the problem is: TCP runs a discovery first, to find out who it should send the INFO message to. The issue is that MERGE3 waits for `(min_interval+max_interval)/2` ms, which in the test is ~5 seconds. This delays the sending of INFO messages by 5 seconds. 

UDP skips this step as it can simply send the INFO message as a _multicast message_.

Because we're getting INFO messages from fewer members within the `check_interval` timeout in `MERGE3`, multiple merges are needed.

belaban:
Hmm, methinks a reactive approach is better suited for this kind of issue: MERGE3 should send down a `FIND_MBRS` event and pass a ref to a lambda function which is invoked on every response. The lambda function simply sends the INFO message to that member. This is better than waiting for 5+ seconds, and doesn't block the thread. Creating a JIRA...

belaban:
I created https://issues.jboss.org/browse/JGRP-2281 to fix this.

rhusar:
Great, thanks, looks like we are moving. I would still like to point out that, even with UDP:
* the minimum time to merge is still at least `max_interval * 1.6`
* the maximum time to merge is also `max_interval * 1.6` while it would seem from the user perspective the maximum time for the merge would typically be the `max_interval`

belaban:
Note that there's `check_interval`, which is by default set to `max_interval * 1.6`, but you can also set it yourself. However, if you set it to a value smaller than `max_interval`, chances are you'll miss some `INFO` messages and therefore don't have responses from all members.

PR #394, opened 2018-09-12
JGRP-2286 VERIFY_SUSPECT can return false positives if member is killed and immediately restarted

pferraro:
https://issues.jboss.org/browse/JGRP-2286

### DIFF
--- a/src/org/jgroups/protocols/VERIFY_SUSPECT.java
+++ b/src/org/jgroups/protocols/VERIFY_SUSPECT.java
@@ -87,6 +87,7 @@ public class VERIFY_SUSPECT extends Protocol implements Runnable {
         return timeout;
     }
 
+    @Override
     public Object down(Event evt) {
         switch(evt.getType()) {
             case Event.SET_LOCAL_ADDRESS:
@@ -100,6 +101,7 @@ public class VERIFY_SUSPECT extends Protocol implements Runnable {
         return down_prot.down(evt);
     }
 
+    @Override
     public Object up(Event evt) {
         switch(evt.getType()) {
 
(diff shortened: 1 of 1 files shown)


pferraro:
@belaban Please review.

danberindei:
@belaban @pferraro I'm not sure this fixes anything... the logs attached in https://issues.jboss.org/browse/ISPN-9517 show that the old node is removed from the view just fine

pferraro:
@danberindei is right. So long as VERIFY_SUSPECT sends ARE_YOU_DEAD messages as unicasts, they should be ignored by the transport of the reincarnated member via TP.unicastDestMismatch(...)  @belaban You can revert this.

belaban:
Reverted

belaban:
Hmm, this is not true:
`public boolean unicastDestMismatch(Address dest) {
        return dest != null && !(Objects.equals(dest, local_addr) || Objects.equals(dest, local_physical_addr));
    }`

If dest != local_addr, but dest == local_physical_addr, then the message would _not_ get discarded... not sure why I once added the second part of the condition...

pferraro:
@belaban Would that mean this still be an issue when use_ip_addrs is enabled?

danberindei:
@belaban could it be because some protocols (maybe discovery?) need to send messages to a particular physical address? I would expect this kind of usage to be quite limited though, and the protocols that use physical addresses should have some kind of protection against reincarnation themselves.

OTOH I'm not sure `VERIFY_SUSPECT.use_icmp` could ever work... any time the JVM crashes but the machine stays up, the node is not going to be removed from the view. If another node is started on the same machine, it won't remove the old node from the view either.

belaban:
@danberindei Yes, you're right: TCPPING needs to send messages to physical addresses when iterating through `initial_hosts`

belaban:
OK @pferraro Have you come up with a reproducer?
So the current logic does _not_ discard a message if the UUID is the same as the local address OR the destination is a physical address and the local physical address is the same. This means reincarnation is possible *if* the destination is a physical address and it equals the local physical address.
However, this should *only* be possible in TCPPING. That's the only case where we ever send physical addresses rather than UUIDs in a message.

PR #410, opened 2018-12-05
In FILE_PING, ensure that coordinator creates a file for itself if none exists

nsawadsky:
This is a fix for https://issues.jboss.org/browse/JGRP-2288. The fix handles an edge case that can arise when remove_all_files_on_view_change is set to true. In a network partition, a view change in a separate partition may cause the file for this partition's coordinator to be removed. Under specific circumstances, as described in JGRP-2288, the subclusters may then never merge, even after the partition is resolved. 

I've done this fix on the 3.6 branch as that is the branch our product is currently using. I can generate a pull request for master as well, or instead, if that's preferred.

### DIFF
--- a/src/org/jgroups/protocols/FILE_PING.java
+++ b/src/org/jgroups/protocols/FILE_PING.java
@@ -136,7 +136,14 @@ public class FILE_PING extends Discovery {
                     ; // use case #1 if we have predefined files: most members join but are not coord
             }
             else {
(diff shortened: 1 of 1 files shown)


nsawadsky:
Closing in favour of https://github.com/belaban/JGroups/commit/a2ea347d793c1c84f7b8c0db0f485b8223889afb

PR #416, opened 2019-01-08
The timer field is used as a flag to exit the thread's while loop

rvprg:
This is regarding point #2 raised here: https://issues.jboss.org/browse/JGRP-2287

The timer field is used as a flag to exit the thread's while loop but this field's visibility isn't guaranteed. The field should be volatile.

Point #1 seems to be OK to me, as all those methods use locks internally, and the loop should behave as expected.


belaban:
Thanks for the patch! I'll try to address the other 2 issues mentioned in JGRP-2287, too.

PR #423, opened 2019-04-01
JGRP-2242 Narrow set of thrown exceptions in Streamable and Marshaller method signatures.

pferraro:
https://issues.jboss.org/browse/JGRP-2242

Now that 4.0.x was branched, perhaps this is ready for inclusion?

### DIFF
--- a/src/org/jgroups/AnycastAddress.java
+++ b/src/org/jgroups/AnycastAddress.java
@@ -71,6 +71,7 @@ public class AnycastAddress implements Address, Constructable<AnycastAddress> {
         }
     }
 
+    @Override
     public int serializedSize() {
         if (destinations == null) {
             return Global.INT_SIZE;
@@ -124,12 +125,12 @@ public class AnycastAddress implements Address, Constructable<AnycastAddress> {
     }
 
     @Override
-    public void writeTo(DataOutput out) throws Exception {
+    public void writeTo(DataOutput out) throws IOException {
         Util.writeAddresses(destinations, out);
     }
 
     @Override
-    public void readFrom(DataInput in) throws Exception {
+    public void readFrom(DataInput in) throws IOException, ClassNotFoundException {
         destinations=Util.readAddresses(in, ArrayList::new);
     }
 
--- a/src/org/jgroups/MergeView.java
+++ b/src/org/jgroups/MergeView.java
@@ -7,6 +7,7 @@ import org.jgroups.util.Util;
 
 import java.io.DataInput;
 import java.io.DataOutput;
+import java.io.IOException;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
@@ -101,8 +102,8 @@ public class MergeView extends View {
         return sb.toString();
     }
 
-
-    public void writeTo(DataOutput out) throws Exception {
+    @Override
+    public void writeTo(DataOutput out) throws IOException {
         super.writeTo(out);
 
         // write subgroups
@@ -130,7 +131,8 @@ public class MergeView extends View {
         }
     }
 
-    public void readFrom(DataInput in) throws Exception {
+    @Override
+    public void readFrom(DataInput in) throws IOException, ClassNotFoundException {
         super.readFrom(in);
         short len=in.readShort();
         if(len > 0) {
@@ -154,6 +156,7 @@ public class MergeView extends View {
         }
     }
 
+    @Override
     public int serializedSize() {
         int retval=super.serializedSize();
         retval+=Global.SHORT_SIZE; // for size of subgroups vector
--- a/src/org/jgroups/Message.java
+++ b/src/org/jgroups/Message.java
@@ -7,6 +7,7 @@ import org.jgroups.util.*;
 
 import java.io.DataInput;
 import java.io.DataOutput;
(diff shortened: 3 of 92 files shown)


belaban:
What's the rationale again for changing almost all of JGroups? :-)

pferraro:
@belaban This removes the need for consumers of these methods to consistently handle a non-specific java.lang.Exception - particularly useful when adapting to other marshalling frameworks.

belaban:
Do you have an example (code)?

pferraro:
Off the top of my head:
https://github.com/wildfly/wildfly/blob/master/clustering/server/src/main/java/org/wildfly/clustering/server/group/AddressSerializer.java#L49
https://github.com/wildfly/wildfly/blob/master/clustering/server/src/main/java/org/wildfly/clustering/server/group/AddressSerializer.java#L61
https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsAddress.java#L69
https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsAddress.java#L78
https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTopologyAwareAddress.java#L131
https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTopologyAwareAddress.java#L140

belaban:
Hmm... so your code throws an IOException... But [1] is inconsistent: it throw IOException on write() and also on read(), but other code throws {{IOException, ClassNotFoundException}} on read().

I'll think about this

[1] https://github.com/wildfly/wildfly/blob/master/clustering/server/src/main/java/org/wildfly/clustering/server/group/AddressSerializer.java#L60

pferraro:
These are pretty standard exception types to throw when reading/writing to/from an java.io.ObjectInput/ObjectOutut stream. See:
https://docs.oracle.com/javase/8/docs/api/java/io/Externalizable.html
https://github.com/wildfly/wildfly/blob/master/clustering/marshalling/api/src/main/java/org/wildfly/clustering/marshalling/Externalizer.java
https://github.com/infinispan/infinispan/blob/master/commons/src/main/java/org/infinispan/commons/marshall/Externalizer.java
https://github.com/jboss-remoting/jboss-marshalling/blob/master/api/src/main/java/org/jboss/marshalling/Externalizer.java

This is because ObjectInput.readObject() itself throws both an IOException *or* a ClassNotFoundException, if the class descriptor from the stream cannot be loaded.
https://docs.oracle.com/javase/8/docs/api/java/io/ObjectInput.html#readObject--
Whereas ObjectOutput.writeObject(...) only ever throws an IOException (i.e. it cannot be the case that the class was not found, if we already have an instance of the class)
https://docs.oracle.com/javase/8/docs/api/java/io/ObjectOutput.html#writeObject-java.lang.Object-

belaban:
OK, so both methods will only throw an IOException.
I'll change this in 4.1.0

belaban:
I'll take a look, too. I'm currently busy with the port to Quarkus, but will take a look early next week.

belaban:
I don't like having to add a {{ClassNoutFoundException}} to the readXXX() methods. As in your examples above, I'd like to add this exception only to methods (e.g. {{Util.readAddress()}}) that possibly load classes.
Adding this to {{Streamable.readFrom()}} is not desired.
WDYT?

belaban:
> These are pretty standard exception types to throw when reading/writing to/from an java.io.ObjectInput/ObjectOutut stream. See:
> https://docs.oracle.com/javase/8/docs/api/java/io/Externalizable.html
> https://github.com/wildfly/wildfly/blob/master/clustering/marshalling/api/src/main/java/org/wildfly/clustering/marshalling/Externalizer.java
> https://github.com/infinispan/infinispan/blob/master/commons/src/main/java/org/infinispan/commons/marshall/Externalizer.java
> https://github.com/jboss-remoting/jboss-marshalling/blob/master/api/src/main/java/org/jboss/marshalling/Externalizer.java
> 
> This is because ObjectInput.readObject() itself throws both an IOException _or_ a ClassNotFoundException, if the class descriptor from the stream cannot be loaded.
> https://docs.oracle.com/javase/8/docs/api/java/io/ObjectInput.html#readObject--
> Whereas ObjectOutput.writeObject(...) only ever throws an IOException (i.e. it cannot be the case that the class was not found, if we already have an instance of the class)
> https://docs.oracle.com/javase/8/docs/api/java/io/ObjectOutput.html#writeObject-java.lang.Object-

While that is correct, in most cases, JGroups does _not_ read/write Objects, but simply - primitive - types.

pferraro:
> I'd like to add this exception only to methods (e.g. Util.readAddress()) that possibly load classes.

That would be a welcome improvement.

belaban:
OK. I'll accept your PR, but will make additional modifications, as not all methods (e.g. in Util) have the correct exceptions.

belaban:
Pushed to master. Let me know if this works for you

pferraro:
That works. Thanks!

PR #431, opened 2019-07-15
JGRP-2359 Route list can be empty

pruivo:
https://issues.jboss.org/browse/JGRP-2359

See JIRA for details. 

I decided to do a copy to perform the changes from the view change. However, I'm open to suggestions.

Also, the Relay2Test is failing (for me) in master and I was unable to test if this breaks anything :)


pruivo:
ps. needs backport to 4.0.x :)

belaban:
Why the need for the copy? We use a ConcurrentHashMap for `routes`; if you want to do a route and the list is empty, can't you handle it (= an empty list)? This means a route was removed in `viewAccepted()` while a route() was going on...

belaban:
Oops, I see this is for the value (`List<Route>`)! OK...

pruivo:
@belaban Thanks!

PR #432, opened 2019-08-07
JGRP-2366 Log UDP network interface and port range

diegolovison:


diegolovison:
@bertrama done. moved to `dumpSocketInfo`

PR #433, opened 2019-08-13
Change method name 'size' to 'getResponsesCount'

pdhung3012:
This class is used to represent ResponseCollector.  This method named 'size' is to get the number of responses. Thus, the method name 'getResonsesCount' is more intuitive than 'size'.


belaban:
I disagree; I think `size()` is perfectly fine! Also, this might break existing code...

PR #434, opened 2019-08-16
Update maven version to SNAPSHOT

rhusar:
Development branches like `master` need to use `-SNAPSHOT` version suffix not to spoil local maven repositories.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
     <artifactId>jgroups</artifactId>
     <packaging>bundle</packaging>
     <name>JGroups</name>
-    <version>4.1.3.Final</version>
+    <version>4.1.4.Final-SNAPSHOT</version>
     <url>http://www.jgroups.org</url>
 
     <properties>


rhusar:
@belaban This is even worse... the 4.1.3.Final is released in Jira but the tag is missing ;-(

belaban:
I can't change the version just yet, as I had an issue uploading the 4.1.3.Final artifcat to maven. I'm trying again today, with the help of David...

belaban:
I *did* create a tag JGroups-4.1.3.Final and push it to GIT!

rhusar:
Ok, I see the tag now, must have been a race condition. :-) 

Regardless, master must be *always* SNAPSHOT version, otherwise people push their changes on rebased master branch as final GA versions that won't get overriden. That's why there are things like maven release plugin that makes all of this hassle free https://maven.apache.org/maven-release/maven-release-plugin/ – also releases are done from a tag, not from a branch, so it doesn't matter what revising is master on.

belaban:
Care to contribute a release plugin change to the POM? :-)

belaban:
OK, I set master to 4.1.4-SNAPSHOT and will release from the release-4.1.3 branch (once David returns from his PTO).

PR #441, opened 2019-09-23
Adding IV support to SYM_ENCRYPT and ASYM_ENCRYPT protocols (JGRP-2386).

nsawadsky:
This change allows JGroups to support symmetric encryption algorithms that require an initialization vector (IV). An important example is "AES/CBC/PKCS5Padding" which uses the CBC cipher mode. 

The main changes that were required are as follows:

- Added a sym_iv_length field to Encrypt. This must be set to a non-zero value when using a cipher mode that requires an IV.
- Added an iv field to the EncryptHeader (which may be null if the cipher mode does not require an IV).
- Since the IV is passed when the Cipher is initialized and the IV is created per-message, cipher initialization is now deferred until the message is being encrypted/decrypted.
- Since the key is passed to the Cipher.init() method and init() is deferred, we now store the Key rather than the Cipher in the key_map field of Encrypt.
- DH_KEY_EXCHANGE is mostly unchanged. So it does not support using an IV-based cipher to encrypt the shared symmetric key. This seems okay as a fresh DH session key is generated for each key exchange, so the ciphertext will always be different. This saves us having to add an iv field to the DhHeader. 
- All the SYM_ENCRYPT and ASYM_ENCRYPT tests have been extended with subclasses that run the same tests using "AES/CBC/PKCS5Padding".

### DIFF
--- a/src/org/jgroups/protocols/ASYM_ENCRYPT.java
+++ b/src/org/jgroups/protocols/ASYM_ENCRYPT.java
@@ -327,7 +327,7 @@ public class ASYM_ENCRYPT extends Encrypt<KeyStore.PrivateKeyEntry> {
                 log.trace("%s: asking %s to fetch the shared group key %s via an external key exchange protocol (srv=%s)",
                           local_addr, encr_msg.getDest() == null? "all members" : encr_msg.getDest(),
                           Util.byteArrayToHexString(sym_version), srv);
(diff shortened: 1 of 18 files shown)


belaban:
LGTM. The 'encrypt' task passed fine, now running the entire testsuite... stay tuned.

belaban:
@nsawadsky Do you think a small addition to the manual might make sense?

belaban:
@nsawadsky Thanks for your contribution!

nsawadsky:
Thanks Bela! I will open another pull request for the manual updates.

PR #442, opened 2019-10-03
JGRP-2388 DNS_PING#destroy could yield NPE hiding the root cause

rhusar:
https://issues.jboss.org/browse/JGRP-2388


rhusar:
@belaban  I ll remember next time :-)

belaban:
@rhusar You da man! :-)

PR #443, opened 2019-10-05
Switching to AES/CBC/PKCS5Padding in manual and conf examples.

nsawadsky:
I updated the examples in the manual and in the conf folder to use AES/CBC/PKCS5Padding.


belaban:
Just to let you know: I rebased your changes onto 5.0, and this works perfectly!

nsawadsky:
Good to hear!

PR #444, opened 2019-11-01
JDG-3264 Message from a non-member causes FD_ALL to continually suspect it

diegolovison:
https://issues.jboss.org/browse/JDG-3264
JDG-3264 Message from a non-member causes FD_ALL to continually suspect it


diegolovison:
@belaban it is missing FD_ALL2 but I would like first to know what is your opinion.
Commenting you fix the test will fail.

belaban:
@diegolovison Hi Diego,
I don't see (1) what this PR is testing and (2) it is *not* testing the reception of a message by a non-member!

PR #445, opened 2019-11-05
JGRP-2394 Provide ObjectName prefix also for unregistrations

anistor:
https://issues.jboss.org/browse/JGRP-2394


belaban:
Thanks, applied, committed and pushed

PR #446, opened 2019-11-13
[JGRP-2399] Problem with loading resources from the bundle root (/) in OSGI environment

stalyanov:
[JGRP-2399] Problem with loading resources from the bundle root (/) in OSGI environment

+moved resources to META-INF/resources
+pushed version to 3.6.20.Final

### DIFF
--- a/conf/JGROUPS_VERSION.properties
+++ b/conf/JGROUPS_VERSION.properties
@@ -1,5 +1,5 @@
 
 ## Defines the JGroups version
 ## This file should be the *only place* to change when changing the version number
-jgroups.version=3.6.20-SNAPSHOT
+jgroups.version=3.6.20.Final
 jgroups.codename=Tourmalet
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
     <artifactId>jgroups</artifactId>
     <packaging>bundle</packaging>
     <name>JGroups</name>
-    <version>3.6.20-SNAPSHOT</version>
+    <version>3.6.20.Final</version>
     <url>http://www.jgroups.org</url>
 
     <properties>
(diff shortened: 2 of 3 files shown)


belaban:
I'll try this out tomorrow, thanks.
Question: this is against branch 3.6 only. Don't you want to apply this against upstream, too? Otherwise, your changes won't be in more recent releases!

belaban:
OK, so this fails as META-INF/resources is not found. This is not a known location; who creates this?

stalyanov:
> OK, so this fails as META-INF/resources is not found. This is not a known location; who creates this?

@belaban Are you experiencing build or runtime error ? Can you share some more details ?
Resources are copied from `/conf` to `/META-INF/resources` by the maven resource plugin

```xml
<resources>
            <resource>
                <directory>conf</directory>
                <targetPath>META-INF/resources</targetPath>
                <includes>
                    <include>*.xml</include>
                    <include>*.properties</include>
                </includes>
```

I was using maven 3.3.9 to build it - the usual `mvn clean install`

belaban:
I don't use maven to build artifacts, but ant&ivy.

belaban:
I also don't like having to specify META-INF/resources explicitly; this should be on the classpath (or modulepath)!

stalyanov:
Hi @belaban ,I see your point ... and my problem is that I need to integrate jgroups 3.6.x as part of a legacy osgi project ... and I am really not sure what are my other options beside moving resources out of the jar root ... I will appreciate if you have any suggestions on this ?

belaban:
I don't see what this changes: instead of `meta-inf` you now use `resources`. This is a dir that's not created by default...

stalyanov:
@belaban , ok, I will move resource directory creation and file copy in `ant`  and I will provide changes for the 4.x version as well...

belaban:
Wait! I don't want to load resources from a special directory within the JAR. If those resources are not on the classpath, then that's an issue with configuration, IMO.
I suggest create your own JAR, based on a release tag (e.g. 3.6.18.Final).
I'm also thinking about dropping OSGi support...

PR #466, opened 2019-11-24
Replace org.testng.annotations.Test#parallel with org.testng.annotati…

rhusar:
…ons.Test#singleThreaded

This should do the trick for the new TestNG


rhusar:
As per javadoc https://javadoc.io/doc/org.testng/testng/6.9.4/org/testng/annotations/Test.html

belaban:
Thanks, Rado!

belaban:
Also removed singleThreaded=false expression (redundant)

rhusar:
Makes sense.

PR #476, opened 2020-01-03
JGRP-2429 MsgStats num_bytes_sent should be incremented by the Message::size

diegolovison:
https://issues.redhat.com/browse/JGRP-2429


diegolovison:
Should we change `incrNumBytesReceived` in the same PR ?

belaban:
See my comment in the JIRA

PR #478, opened 2020-01-14
JGRP-2434 Support multiple TShark fields parser

diegolovison:
https://issues.redhat.com/browse/JGRP-2434

### DIFF
--- a/conf/jgrp.yaml
+++ b/conf/jgrp.yaml
@@ -21,8 +21,8 @@ spec:
       containers:
       - image: belaban/jgrp
         name: jgrp
-        command: ["bash"]
-        args: ["org.jgroups.demos.Chat -props udp.xml -nohup"]
+        command: ["chat.sh"]
+        args: ["-props udp.xml -nohup"]
         env:
           - name: DNS_QUERY
             value: "jgrp.default.svc.cluster.local"


belaban:
How is TSharkDumpUtil supposed to work?
`tshark -q -Tfields -e data -e frame.time_epoch udp and dst 232.5.5.5 > bela.dump` and then
`java org.jgroups.util.TSharkDumpUtil -file ~/bela.dump  -tshark-fields frame.time_epoch` doesn't work:
```
Exception in thread "main" java.lang.NumberFormatException: For input string: "2800020002020302f3565f977f55fbfed4d53394a7cf09180002000500350101000464726177010039003c000464726177000000210102f3565f977f55fbfed4d53394a7cf091800010001411004c0a80169d5adffff"
	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.base/java.lang.Long.parseLong(Long.java:692)
	at java.base/java.lang.Long.valueOf(Long.java:1144)
	at org.jgroups.util.TSharkDumpUtil.main(TSharkDumpUtil.java:97)
```

diegolovison:
@belaban I created a separated commit to show what I changed. Once we are ok, I can squash and then we can merge.

In order to avoid the issue that you mentioned, I added a validation and an help method.
`-tshark-fields: -e arguments separated by comma in the same order of those provided in tshark command`

diegolovison:
Moved `TSharkDumpUtil` to `tests/other/org/jgroups/tests/TSharkDumpUtil.java` and squashed the commits

belaban:
Why a new class? Can't this be added to ParseMessages?

diegolovison:
To `ParseMessages` support more than one field we will need to change https://github.com/belaban/JGroups/blob/ab88e8e371918abcc4e66c06df7880c51d1a5fda/src/org/jgroups/util/Util.java#L1479 to read `dis.read()` instead of `dis.readShort()`.

When running tshark with more than one field the separator will be `\t`.

Once we make this change, we will need to introduce the `String[] tSharkFields` in https://github.com/belaban/JGroups/blob/master/src/org/jgroups/util/Util.java#L1450

To avoid changing all those stuffs in the `Util` class I decided to read the line and then split by `\t`. See https://github.com/belaban/JGroups/pull/478/files#diff-a57199f02cf6030e17e4713506898fb3R98

Also, the `Util::parse` is used in `GossipRouter`

diegolovison:
@belaban how about a new project https://github.com/jgroups-extras/jgroups-tshark-parser ?
Because I would like to add more parser like: FD_ALL from JGroups logs compared with TShark FD_ALL dumps.

rhusar:
@diegolovison BTW if you are going to start a new repo in jgroups-extras, then lets make it general tooling package, rather than a single tool repo.

belaban:
I still don't see why this couldn't be done as part of JGroups... creating a new repo should only be considered if this functionality doesn't belong into JGroups itself

diegolovison:
In this case, if the team is OK we can merge it.

belaban:
To do that, you'd have to reduce the size of your PR and explain why we have to use a separate parsing method... :-)

diegolovison:
I didn't understand why it was closed when I did a force push.
I will open a new one

rhusar:
@diegolovison It's automatically closed when HEAD of the branch in the PR is the same as HEAD of the target branch. In other words, you messed up the rebase :)

PR #500, opened 2020-07-01
JGRP-2486 FD Monitor gets stuck on TransferQueueBundler

LBrandl:
The FD Monitor was split into a heartbeat sender task and a timeout check task, similar to FD_ALL and FD_ALL2. This prevents the timeout check in the FD protocol from being block by the transport.
Ticket: https://issues.redhat.com/browse/JGRP-2486.

### DIFF
--- a/src/org/jgroups/protocols/FD.java
+++ b/src/org/jgroups/protocols/FD.java
@@ -94,8 +94,12 @@ public class FD extends Protocol {
 
     // task that performs the actual monitoring for failure detection
     @GuardedBy("lock")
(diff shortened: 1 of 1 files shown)


belaban:
Lukas, I've merged this into 4.0.22. However, if you plan to move to another 4.x version, then this should also be applied upstream, ie. in the 4.x branch.
Note that I've deprecated `FD` in 5.0

belaban:
I'm going to release a 4.0.23 version, as a one-time exception for you.

belaban:
Actually, 4.0.24.Final

PR #501, opened 2020-07-02
JGRP-2486 FD Monitor gets stuck on TransferQueueBundler

LBrandl:
The FD Monitor was split into a heartbeat sender task and a timeout check task, similar to FD_ALL and FD_ALL2. This prevents the timeout check in the FD protocol from being block by the transport.
Ticket: https://issues.redhat.com/browse/JGRP-2486.


belaban:
Thanks! Applied

PR #504, opened 2020-07-15
JGRP-2489 TcpConnection.Receiver#run should use logger to handle exce…

rhusar:
…ption rather than rely on the UncaughtExceptionHandler

https://issues.redhat.com/browse/JGRP-2489


rhusar:
PR for 4.x is https://github.com/belaban/JGroups/pull/505

PR #514, opened 2020-09-30
JGRP-2505: (4.x) change dns_query to support a comma-separated list of queries

savitha-ping:
It turns out that our application team is still on JGroups 4.x and not updating to 5.x for the next release. Would you be willing to roll this into the next 4.x release? Would really appreciate it! Thanks :)

In multi-region deployments, it'd be useful to combine addresses from multiple DNS queries to form the cluster. This simple change makes the "dns_query" parameter to DNS_PING a comma-separated list of queries as opposed to a single query like it is currently.


belaban:
Absolutely, thanks for the PR!

belaban:
@savitha-ping Note that I just released 4.2.6.Final. It should be available in maven central in 1 hr. Cheers,

PR #520, opened 2020-11-06
Clarify keystore type with SYM_ENCRYPT

oraNod:
A couple things in encryption docs I found slightly misleading.

- SYM_ENCRYPT and ASYM_ENCRYPT need to go above NAKACK2.
- You can use other keystore formats with SYM_ENCRYPT with the keystore_type config attribute.

### DIFF
--- a/doc/manual/protocols.adoc
+++ b/doc/manual/protocols.adoc
@@ -1779,10 +1779,9 @@ _entire message_ into the payload of a new message that's then encrypted and sen
 This is done by SYM_ENCRYPT. The configuration includes mainly attributes that define the keystore, e.g. `keystore_name`
 (name of the keystore, needs to be found on the classpath), `store_password`, `key_password` and `alias`.
 
-SYM_ENCRYPT uses store type JCEKS (for details between JKS and JCEKS see here), however `keytool` uses JKS, therefore
-a keystore generated with keytool will not be accessible.
+SYM_ENCRYPT uses store type JCEKS by default. To use a keystore in another format, use the `keystore_type` attribute.
 
-To generate a keystore compatible with JCEKS, use the following command line options to keytool:
+To generate a keystore in JCEKS format with keytool, use the following command line options:
 
 ----
 keytool -genseckey -alias myKey -keypass changeit -storepass changeit  -keyalg Blowfish -keysize 56 -keystore defaultStore.keystore -storetype  JCEKS


belaban (changes requested):
Hi @oraNod 
No, the encryption protocols need to go _below_ NAKACK2!! Take a look at the sample configs.

oraNod:
> Hi @oraNod
> No, the encryption protocols need to go _below_ NAKACK2!! Take a look at the sample configs.

Thanks @belaban ! I thought you were going to hit on this. It's something I find slightly confusing. Bear with me if this is something that should be obvious of self-explanatory. I understand that the ordering matters so that NAKACK2 is also encrypted.

When I look at the example configs, it appears to me that the *_ENCRYPT protocols occur in the stack before NAKACK2. This is why it lead me to believe there was a typo (before = above). With the Infinispan docs, we use stack.combine="INSERT_BEFORE" like this:

```
<ASYM_ENCRYPT ... stack.combine="INSERT_BEFORE" stack.position="pbcast.NAKACK2"/>
```

Maybe I'm thinking about the order of the stack in the wrong way after looking at Infinispan configs.

I'll revert the changes in this PR but wanted to explain. Hopefully it makes sense and thank you for the corrections.

belaban:
Yes, this is confusing! I always talk about protocols wrt their position in the running stack, *not* the position in the config.
Note that I once wanted to change the config, to reflect the actual running stack, but this caused an uproar, as all configs would have had to be changed... :-)

oraNod:
I see. Thanks for explaining the background there! I've changed it back.

belaban (changes requested):
Hi @oraNod 
I see a lot of whitespace changes.
Please don't update protocols.adoc *directly*! This is generated code, by looking at the protocols and grabbing all of the attributes. PropertiesToAsciidoc is used here...
So you need to edit the code directly. If you want, I can show you how to do this and then generate protocols.adoc

oraNod:
@belaban Thanks again for explaining how JGroups docs are built. I updated the PR to remove the commit with all those unintentional whitespace changes (it was my editor). 

I also looked into SYM_ENCRYPT.java to make sure I applied changes in the correct place and then did a local build. It seems like the changes I made are outside the generated docs.

belaban:
Thx Done!

PR #523, opened 2020-12-08
JGRP-2434 Parse epoch TShark field

diegolovison:
Using the following file
```
1607434722.405997000	2880020000020006021eeb06d15c9fa0fab56d3a343ef9ada800020029004e0101010039003c000470696e670000000101
1607434723.341212000	2880020002020201021eeb06d15c9fa0fab56d3a343ef9ada800020029004e0401010039003c000470696e67
1607434726.383524000	2880020000020006024ecfa5210cee1a57e133bb83e9cd70c700020029004e0101020039003c000470696e670000000101
1607434727.370760000	2880020002020201024ecfa5210cee1a57e133bb83e9cd70c700020029004e0401020039003c000470696e67
```

You can run the `org.jgroups.tests.ParseMessages` with `-parse_time_epoch -file /path/to/example.dump`

### DIFF
--- a/src/org/jgroups/util/Util.java
+++ b/src/org/jgroups/util/Util.java
@@ -16,6 +16,7 @@ import org.jgroups.protocols.relay.SiteUUID;
 import org.jgroups.stack.IpAddress;
 import org.jgroups.stack.Protocol;
 import org.jgroups.stack.ProtocolStack;
+import org.jgroups.tests.ParseMessages;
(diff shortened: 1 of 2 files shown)


belaban:
Hi Diego
this code is too intrusive. Please make it simpler; for instance, Util.parse() can be given an additional argument (`parse_epoch`) which reads the epoch time before reading the rest of the line. Similar to the `tcp` argument.

PR #524, opened 2020-12-29
JGRP-2503 Extend DataInput & DataOutput to avoid copy of byte[]

leonchen83:
`ByteArrayDataInputStream` add an extra method `readBuffer(int)` that avoid byte array copy in `readFully(byte[])` method. this method also can used in user define Streamable interface.
Ticket [https://issues.redhat.com/browse/JGRP-2503](https://issues.redhat.com/browse/JGRP-2503)
Signed-off-by: Baoyi Chen <chen.bao.yi@qq.com>

### DIFF
--- a/src/org/jgroups/NioMessage.java
+++ b/src/org/jgroups/NioMessage.java
@@ -3,6 +3,7 @@ package org.jgroups;
 
 
 import org.jgroups.util.ByteArray;
+import org.jgroups.util.ByteArrayDataInputStream;
 import org.jgroups.util.Util;
 
 import java.io.DataInput;
@@ -218,15 +219,26 @@ public class NioMessage extends BaseMessage {
         int len=in.readInt();
         if(len < 0)
             return;
-        // unfortunately, we cannot create a ByteBuffer and read directly into it from an input stream (no such API)
-        byte[] tmp=new byte[len];
-        in.readFully(tmp, 0, tmp.length);
-        buf=createBuffer(tmp, 0, tmp.length);
+        if(in instanceof ByteArrayDataInputStream) {
+            ByteArrayDataInputStream bin=(ByteArrayDataInputStream) in;
+            buf=createBuffer(bin.readBuffer(len));
+        } else {
+            // fallback
+            byte[] tmp=new byte[len];
+            in.readFully(tmp, 0, tmp.length);
+            buf=createBuffer(tmp, 0, tmp.length);
+        }
     }
 
     protected ByteBuffer createBuffer(byte[] array, int offset, int length) {
         return use_direct_memory_for_allocations? Util.wrapDirect(array, offset, length)
           : ByteBuffer.wrap(array, offset, length);
+    }
+
+    protected ByteBuffer createBuffer(ByteBuffer buffer) {
+        if(!use_direct_memory_for_allocations) 
+            return buffer;
+        return Util.wrapDirect(buffer.array(), buffer.position(), buffer.remaining());
     }
 
 }
--- a/src/org/jgroups/util/ByteArrayDataInputStream.java
+++ b/src/org/jgroups/util/ByteArrayDataInputStream.java
@@ -15,6 +15,7 @@ public class ByteArrayDataInputStream extends InputStream implements DataInput {
 
     // index of last byte to be read, reading beyond will return -1 or throw EOFException. Limit has to be <= buf.length
     protected final int    limit;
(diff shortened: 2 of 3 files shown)


leonchen83:
if add `readBuffer(int)` to `ByteArrayDataInputStream` user define Streamable that also can use this method to avoid array copy. 
for example
```
public class MyMessage implements SizeStreamable, Constructable<MyMessage> {
	private ByteBuffer payload;
	
	public MyMessage() {
	}
	
	public MyMessage(ByteBuffer payload) {
		this.payload = payload;
	}
	
	@Override
	public Supplier<MyMessage> create() {
		return () -> new MyMessage();
	}
	
	@Override
	public int serializedSize() {
		return payload.remaining();
	}
	
	@Override
	public void writeTo(DataOutput out) throws IOException {
		out.writeInt(payload.remaining()); 
		out.write(payload.array(), payload.position(), payload.remaining());
	}
	
	@Override
	public void readFrom (DataInput in) throws IOException {
		if (in instanceof ByteArrayDataInputStream) {
			// NO COPY
			ByteArrayDataInputStream bin = (ByteArrayDataInputStream) in;
			int len = bin.readInt();
			payload = bin.readBuffer(len);
		} else {
			// COPY
			int len = in.readInt();
			byte[] tmp = new byte[len];
			in.readFully(tmp);
			payload = ByteBuffer.wrap(tmp);
		}
	}
}
```

belaban:
Excellent pull request, thank you!
I've made one minor change and am running the test suite now. Will merge momentarily.
Cheers,

belaban:
I'm having issues with FragTest; except for FRAG, FRAG2, FRAG3 and FRAG4 handle NioMessage fragmentation incorrectly! Good to have unit tests, fixing this now...

belaban:
Oops, bad news: this won't work for UDP and SHARED_LOOPBACK. It may work for TCP though. The reason is that UDP has a fixed buffer (`UDP.PacketReceiver.receive_buf`), into which data is read from the datagram socket. This data is then passed down, for marshalling to ensue. However, if NioMessage creates a ByteBuffer based on an offset/length into `receive_buf`, the data in `receive_buf` will be overwritten on the next read, making the previous read read garbage.
This is different for TCP: it does not read from a fixed buffer, but creates the ByteArrayDataInputStream directly from the Socket. However, I suspect that a socket also maintains some sort of fixed (ring) buffer; perhaps reading small messages off of a socket is ok.
I'm going to close this, sorry! If you have an idea how we can still salvage the elimination of the memory allocation, I'm all ears!

leonchen83:
Hi
could we rollback `NioMesage` but retain change of `ByteArrayDataInputStream`? 
then we can also define `MyMessage` to reduce user define message byte array copy.

belaban:
When do you intend to use `ByteArrayDataInputStream.readBuffer()`?

leonchen83:
only used in `MyMessage .readFrom` and make sure to consume all byte from ByteBuffer.

belaban:
But what's a real-world scenario? MyMessage is basically the same as NioMessage!
Where does the DataInput come from in MyMessage.read()?

leonchen83:
I edited `MyMessage `. if we do not pass `ByteArrayDataInputStream .readBuffer` to outside of `readFrom` like `MyMessage` above. is that work fine？

belaban:
I want to know the use case; where do you use MyMessage (e.g. over ObjectMessage or NioMessage)?
Note that you could also create your own subclass of ByteArrayDataInputStream and add readBuffer to it...

leonchen83:
we define a receiver
protected final class ReceiverImpl implements Receiver {
		
		@Override
		public void receive(final org.jgroups.Message message) {
			final MyMessage m = ((ObjectMessage)message).getObject();
			Object object = m.getObject();
			// do something with object in another thread
		}
		...
		other method...
}
```
register receiver to JChannel
```
JChannel channel = new JChannel(address.getProperties().openStream());
channel.setName(name);
final Receiver receiver = new ReceiverImpl();
channel.setReceiver(receiver);
channel.connect(this.address.getName());
```

use following method send message 
```
channel.send(new ObjectMessage(null, new MyMessage(object))));
```

above is our real-world scenario but more complex.

leonchen83:
if we create our own subclass of ByteArrayDataInputStream and add readBuffer . how can I register my subclass to JGroups?

belaban:
I added readBuffer() to ByteArrayDataInputStream

PR #525, opened 2020-12-30
ClassConfigurator add 2 method with parameter Supplier<T> supplier.

leonchen83:
user define supplier that can allow user do some check dynamically when invoke `readFrom` method so that reduce memory copy.

more details please refer to example

Signed-off-by: Baoyi Chen <chen.bao.yi@qq.com>

### DIFF
--- a/src/org/jgroups/conf/ClassConfigurator.java
+++ b/src/org/jgroups/conf/ClassConfigurator.java
@@ -68,6 +68,10 @@ public class ClassConfigurator {
             add(magic, clazz);
     }
 
+    public static <T> void addIfAbsent(short magic, Class<T> clazz, Supplier<T> supplier) {
+        if(!magicMapUser.containsKey(magic) && !classMap.containsKey(clazz))
+            add(magic, clazz, supplier);
+    }
 
     /**
      * Method to register a user-defined header with jg-magic-map at runtime
@@ -76,6 +80,10 @@ public class ClassConfigurator {
      * @throws IllegalArgumentException If the magic number is already taken, or the magic number is <= 1024
      */
     public static void add(short magic, Class<?> clazz) {
+        add(magic, clazz, null);
+    }
+
+    public static <T> void add(short magic, Class<T> clazz, Supplier<T> supplier) {
         if(magic < MIN_CUSTOM_MAGIC_NUMBER)
             throw new IllegalArgumentException("magic ID (" + magic + ") must be >= " + MIN_CUSTOM_MAGIC_NUMBER);
         if(magicMapUser.containsKey(magic) || classMap.containsKey(clazz))
@@ -97,6 +105,14 @@ public class ClassConfigurator {
 
         if(Constructable.class.isAssignableFrom(clazz)) {
             val=((Constructable<?>)inst).create();
+            inst=((Supplier<?>)val).get();
+            if(!inst.getClass().equals(clazz))
+                throw new IllegalStateException(String.format("%s.create() returned the wrong class: %s\n",
+                                                              clazz.getSimpleName(), inst.getClass().getSimpleName()));
+        }
+
+        if (supplier != null) {
+            val=supplier;
             inst=((Supplier<?>)val).get();
             if(!inst.getClass().equals(clazz))
                 throw new IllegalStateException(String.format("%s.create() returned the wrong class: %s\n",


leonchen83:
Example

1. define Messenger 
```
public class Messenger {
		private JChannel channel;
		private List<String> subChannels = new CopyOnWriteArrayList<>();
		
		public void init() {
			ClassConfigurator.add((short)12346, MyMessage.class, () -> {
				MyMessage message = new MyMessage();
				message.setSubChannels(subChannels);
				return message;
			});
			
			this.channel = new JChannel(address.getProperties().openStream());
			channel.setName(name);
			final Receiver receiver = new ReceiverImpl();
			channel.setReceiver(receiver);
			channel.connect(address.getName());
		}
		
		public void send(String subChannel, byte[] payload) throws Exception {
			channel.send(null, new ObjectMessage(null, new MyMessage(subChannel, payload)));
		}
		
		public void add(String subChannel) {
			subChannels.add(subChannel);
		}
		
		public void remove(String subChannel) {
			subChannels.remove(subChannel);
		}
	}
```

2. define MyMessage
```
public class MyMessage implements SizeStreamable {
		
		private List<String> subChannels;
		
		private String subChannel;
		private byte[] payload;
		
		public String getSubChannel() {
			return subChannel;
		}
		
		public byte[] getPayload() {
			return payload;
		}
		
		public MyMessage() {
		}
		
		public MyMessage(String subChannel, byte[] payload) {
			this.subChannel = subChannel;
			this.payload = payload;
		}
		
		public void setSubChannels(List<String> subChannels) {
			this.subChannels = subChannels;
		}
		
		@Override
		public int serializedSize() {
			return 4 + subChannel.length() + 4 + this.payload.length;
		}
		
		@Override
		public void writeTo(DataOutput out) throws IOException {
			out.writeInt(this.subChannel.length());
			out.writeBytes(this.subChannel);
			out.writeInt(payload.length);
			out.write(payload, 0, payload.length);
		}
		
		@Override
		public void readFrom(DataInput in) throws IOException, ClassNotFoundException {
			int len1 = in.readInt();
			byte[] bytes1 = new byte[len1]; 
			in.readFully(bytes1);
			this.subChannel = new String(bytes1);
			int len2 = in.readInt();
			if (subChannels.contains(subChannel)) {
				byte[] bytes2 = new byte[len2];
				in.readFully(bytes2);
				this.payload = bytes2;
			} else {
				// ignore and avoid readFully copy
				in.skipBytes(len2);
			}
		}
	}
```
3. define ReceiverImpl
```
	public class ReceiverImpl implements Receiver {
		
		@Override
		public void receive(final org.jgroups.Message message) {
			final MyMessage m = ((ObjectMessage)message).getObject();
			byte[] bytes = m.getPayload();
			if (bytes == null) {
				return; // ignore
			} else {
				// do something with bytes
			}
		}
	}
```

4. main method
```
		Messenger messenger = new Messenger();
		messenger.init();
		
		messenger.add("sub-module1");
		messenger.send("sub-module4", new byte[]{0}); // ignore in MyMessage.readFrom
		messenger.add("sub-module2");
		messenger.send("sub-module2", new byte[]{0}); // normal case
```

belaban:
I don't see the point of this PR: MyMessage.writeTo()/readFrom() can use marshalling logic which sends a boolean that determines whether or not to include the payload. Also, you don't IMO need to use the supplier in ClassConfigurator to create a MyMessage: the subchannels can be set by the application after creation of MyMessage...

leonchen83:
in our case we can't set a boolean to check if include payload, because we send multicast request via UDP to every cluster node.
in cluster node side we check by some conditions dynamically to decide whether to process this request. 
if we don't process this request. we don't want to unmarshal this reuqest too. just want to skip payload.

belaban:
You don't need to call reallFully(); you could send the criteria, upon which you decide dynamically whether to call readFull(), based on it (e.g. a String in the example below).

MyMessage.readFrom(DataInput in):

```java
String group=Util.readString(in);
if(/* check criteria on group */) {
    payload=new byte[in.readInt()];
    in.readFully(payload);
}
else {
 // null payload
}
```

leonchen83:
in your code snippets `/* check criteria on group */` means `if (subChannels.contains(subChannel))` in my example
but I don't understand how to set `subChannels`

belaban:
Don't you have a context in which you receive the message? That context can have the subChannels list.
Alternatively, use a MessageFactory and inject subChannels into each MyMessage when received.

leonchen83:
use a self define `MessageFactory` and inject that factory to `TP`. right？

belaban:
Yes: http://www.jgroups.org/manual5/index.html#MessageFactory

leonchen83:
I thought about this plan before.
1. we need create an `MyObjectMessage extends ObjectMessage` and rewrite `readPayload` method
2. we copy `Util.readGenericStreamable` method to `readPayload` and set subchannels
3. define `MyMessageFactory extends DefaultMessageFactory` and register `MyObjectMessage`
4. register MyMessageFactory to `TP`
is that right? It looks a little complicated

PR #532, opened 2021-04-22
Reproducer for JGRP-2554 UUID source address replaced with SiteUUID

danberindei:
https://issues.redhat.com/browse/JGRP-2554

Can't add it as a proper test because it needs an artificial delay in BaseBundler to reproduce.

### DIFF
--- a/src/org/jgroups/protocols/BaseBundler.java
+++ b/src/org/jgroups/protocols/BaseBundler.java
@@ -75,7 +75,7 @@ public abstract class BaseBundler implements Bundler {
         clearMessages();
         count=0;
         try {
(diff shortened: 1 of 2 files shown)


rhusar:
@danberindei I believe we have `tests/byteman` for supporting such test cases that need to modify runtime behavior to reproduce an issue.

danberindei:
@rhusar I'm hoping Bela will find a better way to test it without modifying BaseBundler, if not I'll give Byteman a try.

belaban:
Yes @danberindei  Don't worry about it! If I can reproduce this, then I'll find a way

PR #551, opened 2021-09-08
[4.x] JGRP-2487 TLS Transport

tristantarrant:
https://issues.redhat.com/browse/JGRP-2487

This contains some fixes to make SocketFactories that return SSL sockets work in TCP (UDP and TCP_NIO are not supported).
Additionally the GossipRouter CLI has been enhanced to allow SSL configuration, including the ability to set SNI names.
@pruivo can you please add your review.
Once approved, I'll create a forward port for 5.x

### DIFF
--- a/ivy.xml
+++ b/ivy.xml
@@ -20,6 +20,7 @@
         <dependency org="org.slf4j"                name="slf4j-api"               rev="1.7+"/>
         <dependency org="org.testng"               name="testng"                  rev="6.14.+"/>
         <dependency org="com.beust"                name="jcommander"              rev="1.+"/>
+        <dependency org="org.wildfly.security"     name="wildfly-elytron"         rev="1.16.+"/>
         <!--dependency org="org.sonatype.nexus.ant"   name="nexus-staging-ant-tasks" rev="1.6.3"/-->
     </dependencies>
 
--- a/pom.xml
+++ b/pom.xml
@@ -137,6 +137,12 @@
             <version>4.0.9</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.wildfly.security</groupId>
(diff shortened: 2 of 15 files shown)


belaban:
OK, I'll take a look tomorrow.

tristantarrant:
Note that the `wildfly-elytron` dependency has been added with `test` scope. JGroups will not depend on it :)

ivy.xml
```diff
@@ -20,6 +20,7 @@
         <dependency org="org.slf4j"                name="slf4j-api"               rev="1.7+"/>
         <dependency org="org.testng"               name="testng"                  rev="6.14.+"/>
         <dependency org="com.beust"                name="jcommander"              rev="1.+"/>
+        <dependency org="org.wildfly.security"     name="wildfly-elytron"         rev="1.16.+"/>
```
pruivo, line 23: wildfly-commons is missing. I'm getting NoClassDefFoundError when running `ant all-tests report`
```
org/wildfly/common/Assert

java.lang.NoClassDefFoundError
at org.wildfly.security.x500.cert.SelfSignedX509CertificateAndSigningKey$Builder.setDn(SelfSignedX509CertificateAndSigningKey.java:178)
at org.jgroups.protocols.TLSTest.createSelfSignedCertificate(TLSTest.java:159)
at org.jgroups.protocols.TLSTest.init(TLSTest.java:73)
```

pruivo: looks like `SASL*Test` tests are affected by this as well.

pruivo (approved):

belaban (commented):
It would be nice if GossipRouter had feasible default for tls_provider etc, so if we chose TSL, we would not have to provide all of the detail configuration.

belaban:
OK, so I've merged your changes with slight modifications. Please look at TLS/SASLTest...

PR #556, opened 2021-09-17
[4.x] JGRP-2570 SITE_UNREACHABLE is discarded on originator

pruivo:
SiteMaster was sent back to the originator a message with
SITE_UNREACHABLE but it sets a SiteUUID as destination Address. The
originator discards because it cannot handle this type of Address.

https://issues.redhat.com/browse/JGRP-2570


belaban:
Hi @pruivo can you provide a PR for 5.x as well?
thanks!

PR #569, opened 2021-11-02
[master] Introduce continuous integration testing using GitHub Actions

rhusar:
This GH action will verify compilation using JDK 11 across platforms on master branch and pull requests submitted against it. This will prevent issues like merging uncompilable code, merge conflicts, unresolvable artefacts, etc.

Commits / PRs that have been tested will be rendered on the UI with a green tick next to it. Note that this does not enforce anything (but it is possible to e.g. require passing checks in repo settings if desired).

Going forward, the appropriate tests should be executed, however, they don't seem to be in a runnable state at the moment when executed form maven; also will likely need to account for the environment (e.g. multicast, timing issues).

Here's an example run - https://github.com/rhusar/JGroups/actions/runs/1414565300 - the PR needs to be merged for the CI to actually run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,31 @@
+name: CI
+on:
+  push:
+    branches:
+      - '*'
+  pull_request:
+    branches:
+      - '*'
+jobs:
+  ci:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ ubuntu-latest, windows-latest, macos-latest ]
+        java: [ 11 ]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Set up JDK ${{ matrix.java }}
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java }}
+      - name: Cache local Maven repository
+        uses: actions/cache@v2
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
+      - name: Build with Maven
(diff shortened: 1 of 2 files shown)


belaban:
Same for master?

belaban:
Right, I've not yet tested running the testsuite with maven (I always use ant&ivy). This is on my todo list though (https://issues.redhat.com/browse/JGRP-2499)... If you want to take a look though, be my guest! For IP multicast to run, we should probably always bind to 127.0.0.1 and use 228.8.8.8 as mcast_addr.

rhusar:
@belaban Well, it's not a problem to add a step that runs ant to run the test - here's an example run - https://github.com/rhusar/JGroups/runs/4091354246?check_suite_focus=true - but the tests fail badly, without investigation I am assuming this has to do with binding and mcast. So, there are couple of options:

* add ant step running the tests, let them fail but pass the check (but that gives out false positives)
* add ant step running the tests, but set to fail the build when tests fail which will always fail until we figure out a way to run it reliably in GH environment (but that gives out false negatives)
* leave as is (meaning just compiling with maven and failing only on compilation) and not running the test (but that doesn't detect test regressions)

However, these machines do support multicast, for modcluster I had to create a util that figures out a multicast capable interface and the tests are passing reliably in GH environment, see https://github.com/modcluster/mod_cluster/blob/main/core/src/test/java/org/jboss/modcluster/TestUtils.java#L56-L96

Preferences?

rhusar:
@tristantarrant BTW no more compilation issues on merges ;-)

belaban:
Let's try to fix the testsuite first and use any/ivy for now. This should make is simpler to switch to mvn later.
I suggest setting {{bind_addr}} to {{127.0.0.1}}, however, we can also use custom code to determine the bind address.
Is there a way to debug the GH actions, to see the actual value of {{bind_addr}}?

PR #572, opened 2021-11-10
Add options to MPerf

cjljohnson:
Improve MPerf to track received messages for all receivers, and add
options to ignore local messages and display received per sender


belaban:
Merged into master

PR #573, opened 2021-11-10
JGRP-2531 Early batching of messages

cjljohnson:


belaban:
Committed and merged, thanks!

PR #575, opened 2021-11-25
JGRP-2531 Handle case where finishConnect() returns false even though isConnectable() returned true.

nsawadsky:
Thanks @belaban for merging my previous PR https://github.com/belaban/JGroups/pull/571

This is the backport of the same change to the 4.x branch.


belaban:
Thanks, Nick!

nsawadsky:
Great, we'd like to pick up this change whenever you are able to do a release build of 4.2.18.

PR #589, opened 2022-01-06
Ring Buffer Bundler fixed to write message type before message

cjljohnson:
Since message types were added, a short must be written before messages to indicate the message type for unmarshalling.  This was not done for ring buffer based bundler implementations.


belaban:
Thanks, Chris! Things do run now, but very slowly, will investigate!
Cheers

PR #598, opened 2022-02-07
fix use case DNSResolver for Consul

ruslansennov:
I'm trying to use JGroups with HashiCorp Consul as the DNS engine during service discovery. Now this is impossible, because the Consul resolves the addresses of services in its own namespace (for example, `7f000801.addr.dc1.consul.`) which must be resolved again through `DnsContext`. This PR will fix it.


ruslansennov:
@belaban thank you. Could you make cherry pick to the`4.x` branch?

ruslansennov:
Did you mean `4.2.21`? I'll wait.

PR #604, opened 2022-02-24
Support to generate thread dump in a file

diegolovison:
JGroups generate thread dump in the same log of the server. When dealing with suspected nodes, the logs become big and it is hard to find out the root cause.

This PR allows the developer to set a custom path to generate the thread dump.

### DIFF
--- a/src/org/jgroups/util/ThreadPool.java
+++ b/src/org/jgroups/util/ThreadPool.java
@@ -8,6 +8,10 @@ import org.jgroups.annotations.Property;
 import org.jgroups.conf.AttributeType;
 import org.jgroups.protocols.TP;
 
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
 import java.util.Objects;
 import java.util.concurrent.*;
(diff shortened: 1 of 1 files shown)


src/org/jgroups/util/ThreadPool.java
```diff
@@ -171,10 +176,26 @@ public boolean execute(Runnable task) {
         catch(RejectedExecutionException ex) {
             tp.getMessageStats().incrNumRejectedMsgs(1);
             // https://issues.redhat.com/browse/JGRP-2403
+            String threadDump = Util.dumpThreads();
```
rhusar, line 179: This `org.jgroups.util.Util#dumpThreads` call is in the wrong place. It is expensive so make sure to call it at the code path you are certainly going to use the result.

rhusar (changes requested):

src/org/jgroups/util/ThreadPool.java
diegolovison: Ops. Thanks!

diegolovison:
https://issues.redhat.com/browse/JGRP-2608

PR #621, opened 2022-04-28
JGRP-2614 Non-blocking API for counters

pruivo:
https://issues.redhat.com/browse/JGRP-2614

### DIFF
--- a/src/org/jgroups/blocks/atomic/AsyncCounter.java
+++ b/src/org/jgroups/blocks/atomic/AsyncCounter.java
@@ -1,0 +1,86 @@
+package org.jgroups.blocks.atomic;
+
+import java.util.concurrent.CompletionStage;
+
+/**
+ * An asynchronous counter interface.
+ *
+ * @author Pedro Ruivo
+ * @since 5.2
+ */
+public interface AsyncCounter {
+
+    /**
+     * @return The counter's name.
+     */
+    String getName();
+
+    /**
+     * Gets the current value of the counter.
+     *
+     * @return A {@link CompletionStage} that is completed with the counter's value.
+     */
+    default CompletionStage<Long> get() {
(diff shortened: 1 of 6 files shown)


rhusar:
@pruivo We should have a Jira for this to show up in release notes.

belaban:
@pruivo: you by default implement `get()` as `addAndGet(0)`. Are you sure this is not causing a round-trip to the coordinator, rather than just a local read?

belaban:
I'm adding a unit test

pruivo:
@belaban wondering who wrote that :) 
https://github.com/belaban/JGroups/blob/303bb17254a0d3edddfd7f163316e5e7e9677356/src/org/jgroups/protocols/COUNTER.java#L635-L637

I'll update the PR tomorrow (or this evening)

belaban:
Oh shit, terrible! :-) Yes, please fix it

belaban:
One more issue: on a change (e.g. `incrementAndGet()`), the async API does not update the local counter, so the local counter (stored in hashmap `counters`) looks stale. (Of course, this doesn't happen on the coord). On the next `get()`, the value is updated correctly.

I have to confess that I'm fighting with debugging the async code; I'm not so sure it's a good idea to rewrite the entire `COUNTER` protocol in an async manner...

pruivo:
That's weird and if it is not happening, then it is a bug somewhere.
`sendRequestToCoordinator()` invokes `updateCounter()` when it receives the reply from the coordinator.
Which test are you running?

belaban:
I think `updateCounter()` *is* called, but it has the same version as the updatee, so the new value is not updated...

belaban:
I applied your PR; we can open new PRs/JIRAs for issues found... cheers

pruivo:
thanks @belaban

PR #632, opened 2022-06-29
Fix all broken Jira links

rhusar:
This fixes all broken links that are now spitting out errors:

* http://jira.jboss.com/jira/browse/JGRP
* https://jira.jboss.org/jira/browse/JGRP

and updates the https://issues.jboss.org to the new location saving an unnecessary redirect.


belaban:
This is crazy s**t! Instead of breaking all the links, the old link should have had a forward-rule to the new one!

rhusar:
@belaban Agreed, it's maddening, especially since these URLs are straightforwardly translatable to a redirect.

belaban:
Anyway, thx for fixing this

PR #638, opened 2022-07-19
JGRP-2624 Log MulticastSocket network interface

diegolovison:


src/org/jgroups/protocols/UDP.java
```diff
@@ -513,6 +513,9 @@ protected <T extends UDP> T setNetworkInterface(InetAddress addr, MulticastSocke
             if(s != null && addr != null) {
                 intf=NetworkInterface.getByInetAddress(addr);
                 s.setNetworkInterface(intf);
+                if (log.isDebugEnabled()) {
+                    log.debug(String.format("multicast_socket on %s", intf.getName()));
```
rhusar, line 517: You should use `org.jgroups.logging.Log#debug(java.lang.String, java.lang.Object...)` here instead.

rhusar (changes requested):

PR #643, opened 2022-07-25
JGRP-2626 Fix X509Token Initialization

jfgonca:
- Move the initialization of the X509Token to AUTH.start() method instead of AUTH.init() method since the certificate properties are not defined yet when the init() method is called;
 - Adjust the sample AUTH configurations to include the auth_token. prefix;


belaban:
Thanks!

jfgonca:
Thank you. :)

PR #645, opened 2022-07-27
JGRP-2627 Change multicast address to private network

diegolovison:
https://issues.redhat.com/browse/JGRP-2627

### DIFF
--- a/build.properties.template
+++ b/build.properties.template
@@ -6,7 +6,7 @@
 jgroups.bind_addr=localhost
 jgroups.tcpping.initial_hosts=localhost[7800],localhost[7801]
 jgroups.tunnel.gossip_router_hosts=localhost[12001]
-jgroups.udp.mcast_addr=228.8.8.8
+jgroups.udp.mcast_addr=239.8.8.8
 jgroups.udp.mcast_port=45588
 jgroups.udp.ip_ttl=5
 # use true for IPv6 and false for IPv4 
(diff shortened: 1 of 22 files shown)


belaban:
I'll apply this, but with changes: `224.0.75.75` will remain the disgnostics/probe address, as it in line with mcast addrs for host/routers lookups, such as `224.0.0.1/2`

rhusar:
@belaban Looking at RFC for these ranges, for `224.0.75.75`:

> [6](https://www.rfc-editor.org/rfc/rfc5771.html#section-6).  AD-HOC Blocks (I, II, and III)
> 
>    Addresses in the AD-HOC blocks (including 224.0.2.0 - 224.0.255.255,
>    224.3.0.0 - 224.4.255.255, and 233.252.0.0 - 233.255.255.255) were
>    traditionally used for assignments for those applications that don't
>    fit in either the Local or Internetwork Control blocks.  These
>    addresses MAY be globally routed and are typically used by
>    applications that require small blocks of addressing (e.g., less than
>    a /24 ).  Future assignments of blocks of addresses that do not fit
>    in the Local Network or Internetwork Control blocks will be made in
>    AD-HOC Block III.

The problem here is that these could be globally routed. So I dont think this is an ideal fit.

And for `224.0.0.1`

>  [4](https://www.rfc-editor.org/rfc/rfc5771.html#section-4).  Local Network Control Block (224.0.0/24)
> 
>    Addresses in the Local Network Control Block are used for protocol
>    control traffic that is not forwarded off link.  Examples of this
>    type of use include OSPFIGP All Routers (224.0.0.5) [[RFC2328](https://www.rfc-editor.org/rfc/rfc2328)].

so this seems to be OK.

PR #666, opened 2022-09-28
JGRP-2641 Better invalid magic error

tristantarrant:
https://issues.redhat.com/browse/JGRP-2641

### DIFF
--- a/src/org/jgroups/conf/ClassConfigurator.java
+++ b/src/org/jgroups/conf/ClassConfigurator.java
@@ -132,6 +132,8 @@ public class ClassConfigurator {
                 throw new IllegalStateException(e);
             }
         }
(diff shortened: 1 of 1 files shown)


belaban:
I don't like this solution because users can register their own magic IDs, and this would discard them. Probably move the code into the exception clause, I'll take a look

tristantarrant:
Sorry, I did not know which was the best approach. Look at https://github.com/infinispan/infinispan/discussions/10347 for the reason why I did this

tristantarrant:
Maybe there are some known "wrong magic ids" for which we can provide specific details, e.g. "connecting a RELAY stack to a non-RELAY one", or "connecting a FORK stack to a non-FORK one"

PR #675, opened 2022-10-28
Remove semicolon from if statement

bmccutchon:
It appears that this was only supposed to log when log_collect_msgs is true, but an accidental semicolon resulted in the log statement always running.


belaban:
Wow, good catch! Thanks!

PR #681, opened 2022-11-10
Fix byteman-bmunit to be test dependency.

rhusar:


rhusar:
Hm. Looking at the CI runs, the org.jgroups.util.ProtPerfHelper appears to be a test class in the wrong place. @belaban  Is org.jgroups.util.ProtPerfHelper meant to be directly in jgroups.jar?

belaban:
Reverted

PR #689, opened 2022-11-30
Added check not null for connection JDBC_PING.clearTable

denAbramoff:
if node doesn't have a connection of database when had changed the view - it can't return to cluster after resumption of access

i mean if i have 3 node - a, b, c and ping via jdbc_ping with turn on remove_all_data_on_view_change. Every node on self server. Node A loses connection to database and net - b and c create new view and delete a from table jgroupsping, but node a reckons than b and s left cluster and tries to create self view - but event view_change falls exception npe on method clearTable, because the connection is null. And after resumption of access node a can't to return in cluster b and c

### DIFF
--- a/src/org/jgroups/protocols/JDBC_PING.java
+++ b/src/org/jgroups/protocols/JDBC_PING.java
@@ -111,12 +111,10 @@ public class JDBC_PING extends FILE_PING {
         attemptSchemaInitialization();
     }
 
-
     protected void write(List<PingData> list, String clustername) {
         for(PingData data: list)
             writeToDB(data, clustername, true);
     }
-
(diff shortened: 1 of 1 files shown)


belaban:
Please rebase and squash your commits next time you submit a PR. Also, don't change the formatting, as it's hard to find the real change.
I applied what you changed, please have a look whether this is what you wanted.
Thanks!

denAbramoff:
Yes, thanks!

denAbramoff:
Can you move this change to 3.6.21 version?

belaban:
I'm afraid the 3.6 branch is unmaintained

PR #692, opened 2022-12-15
JGRP-2667: SSL_KEY_EXCHANGE: Support certificate reload and explicit truststore

nsimons:
https://issues.redhat.com/browse/JGRP-2667

### DIFF
--- a/src/org/jgroups/protocols/SSL_KEY_EXCHANGE.java
+++ b/src/org/jgroups/protocols/SSL_KEY_EXCHANGE.java
@@ -6,6 +6,7 @@ import org.jgroups.Global;
 import org.jgroups.View;
 import org.jgroups.annotations.LocalAddress;
 import org.jgroups.annotations.MBean;
+import org.jgroups.annotations.ManagedOperation;
 import org.jgroups.annotations.Property;
 import org.jgroups.conf.AttributeType;
 import org.jgroups.stack.IpAddress;
@@ -20,6 +21,8 @@ import javax.net.ssl.*;
 import java.io.*;
 import java.net.InetAddress;
 import java.security.KeyStore;
+import java.time.Duration;
+import java.time.Instant;
 import java.util.Map;
 import java.util.Objects;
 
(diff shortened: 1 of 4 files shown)


belaban:
Hi Tristan, could you take a look if you have time? Thanks & I owe you a big beer at the next F2F! :-)

belaban:
The test fails with 
```
at java.base/sun.security.pkcs12.PKCS12KeyStore.engineStore(PKCS12KeyStore.java:1214)
	at java.base/sun.security.util.KeyStoreDelegator.engineStore(KeyStoreDelegator.java:189)
	at java.base/java.security.KeyStore.store(KeyStore.java:1411)
	at org.jgroups.protocols.SSL_KEY_EXCHANGE_ReloadTest.saveKeyStoreToFile(SSL_KEY_EXCHANGE_ReloadTest.java:97)
	at org.jgroups.protocols.SSL_KEY_EXCHANGE_ReloadTest.init(SSL_KEY_EXCHANGE_ReloadTest.java:87)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.testng.internal.invokers.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:139)
	at org.testng.internal.invokers.MethodInvocationHelper.invokeMethodConsideringTimeout(MethodInvocationHelper.java:69)
	at org.testng.internal.invokers.ConfigInvoker.invokeConfigurationMethod(ConfigInvoker.java:361)
	at org.testng.internal.invokers.ConfigInvoker.invokeConfigurations(ConfigInvoker.java:296)
	at org.testng.internal.invokers.TestMethodWorker.invokeBeforeClassMethods(TestMethodWorker.java:180)
	at org.testng.internal.invokers.TestMethodWorker.run(TestMethodWorker.java:122)
```

Also, can you make `reloadThreshold` a `@Property`, so users can configure it in their XML file? You may want to convert the attribute to a long (millis) and then create a `Duration` from it. Or you create a `PropertyConverter` to create the `Duration` object directly from the XML...

tristantarrant:
I'll have a look tomorrow

src/org/jgroups/protocols/SSL_KEY_EXCHANGE.java
```diff
@@ -56,6 +58,57 @@ public interface SessionVerifier {
         void verify(SSLSession session) throws SecurityException;
     }
 
+    /**
+     * Helper class that knows how and when to reload the SSLContext.
+     */
+    protected class SSLContextReloader {
```
tristantarrant, line 64: can be made static

src/org/jgroups/protocols/SSL_KEY_EXCHANGE.java
```diff
@@ -82,6 +134,16 @@ public interface SessionVerifier {
     @Property(description="Password to access the keystore",exposeAsManagedAttribute=false)
     protected String          keystore_password="changeit";
 
+    @Property(description="Location of the truststore. Defaults to null, which will use the keystore as a truststore.")
+    protected String          truststore_name;
+
+    @Property(description="The type of the truststore. " +
+      "Types are listed in http://docs.oracle.com/javase/8/docs/technotes/tools/unix/keytool.html")
+    protected String          truststore_type="JKS";
```
tristantarrant, line 142: We should probably change the default to PKCS12 as it is also the JDK default now

tristantarrant (approved):
Two minor suggestions. LGTM otherwise

belaban:
Thx @tristantarrant ! @nsimons: can you incorporate Tristan's and my suggestions, squash and rebase your commits and submit the PR?

src/org/jgroups/protocols/SSL_KEY_EXCHANGE.java
nsimons: Actually we are instantiating the reloader separately for the server and the client ssl context. The current (master) code allows the caller to supply either client or server context (or both), and any missing context is created on the fly. So we want to differentiate these cases. 
For example, if the caller supplies only the server context (client context created by SSL_KEY_EXCHANGE), then we only want to reload the client context. Caller is responsible to reload the server context.

src/org/jgroups/protocols/SSL_KEY_EXCHANGE.java
nsimons: Fixed.

nsimons:
Thanks @belaban and @tristantarrant for the reviews! I have fixed the comments. _reloadThreshold_ is now exposed as a property in the XML and reload can be forced through _probe.sh_.
Would you like to have another look before pushing it in?

belaban:
Thanks @nsimons ! I fucked up, reverted the change, then reverted the revert! :-) Never mind, all's good!

PR #711, opened 2023-04-21
JGRP-2688 Make ProtocolStackConfigurator extend ProtocolHook

pruivo:
https://issues.redhat.com/browse/JGRP-2688

Note: added some useful methods in `ProtocolConfiguration` that aren't related or required to the original JIRA.


wburns:
@belaban is there a reason we can't do this? Without this we have to do some hacky stuff to get component attributes to work from configuration :(

belaban:
@wburns I did apply the PR, but via command line. I should have resolved instead of closed the PR, but a git pull will show you that the result is the same

wburns:
> @wburns I did apply the PR, but via command line. I should have resolved instead of closed the PR, but a git pull will show you that the result is the same

Okay, thanks!

PR #712, opened 2023-04-27
Expose byte[] for ByteArrayDataInputStream

wburns:
Mirror of https://github.com/belaban/JGroups/blob/master/src/org/jgroups/util/ByteArrayDataOutputStream.java#L31-L33


wburns:
Oh whoops I just saw I would need an advance method as well when bytes are "read" outside the scope of the InputStream. I will make a new PR for it. The ByteBufferInputStream already supports this as you can directly modify the underlying ByteBuffer with position(int) method.

PR #721, opened 2023-05-30
do not default to very old protocols, let the jvm pick a secure value

TheGoesen:
TLS 1.2 is fairly old and better alternatives are available. It seems better to rely on the defaults of oracle than to to manually downgrade settings.


belaban:
Excellent! I assume you can live with `sni_matchers` for hostname verification then?

TheGoesen:
I still need to actually test this, I just upgraded everything to 5.2.14 using SYM encrypt... o well

PR #730, opened 2023-06-29
Fix Relay2Header toString() format

pruivo:


belaban (commented):
Hmm, I dont like the change, as I do *not* want to show sites or visited_sites when they're null or empty. Besides, I changed this in the JGRP-2706 branch already

pruivo:
why is that?

belaban:
Only show essential stuff, to not confuse the user

belaban:
Wait for my fix in 2706! Cheers :-)

PR #731, opened 2023-07-31
JGRP-2721 Do not throw exception on corrupted data

pruivo:
The exception closes the connection to the Gossip Router which originates view changes in the other nodes.


belaban:
thx!

PR #732, opened 2023-08-02
JGRP-2722 Gossip Router can corrupt data

pruivo:
Description in JIRA: https://issues.redhat.com/browse/JGRP-2722

* Fix in line 350.
* Another change is in line 544 where I added `NameCache.add(addr, logical_name);`. This avoids printing `null` when it fails to send a message.
* The remaining changes are suggestions by the IDE.

### DIFF
--- a/src/org/jgroups/stack/GossipRouter.java
+++ b/src/org/jgroups/stack/GossipRouter.java
@@ -293,15 +293,14 @@ public class GossipRouter extends ReceiverAdapter implements ConnectionListener,
 
         switch(type) {
             case REGISTER:
-                DataInput in=new ByteArrayDataInputStream(buf);
-                handleRegister(sender, in);
+                handleRegister(sender, new ByteArrayDataInputStream(buf));
                 break;
 
             case MESSAGE:
                 // we already read the type, now read group and dest (minimal info required to route the message)
                 // this way, we don't need to copy the buffer
                 try {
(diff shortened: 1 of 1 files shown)


src/org/jgroups/stack/GossipRouter.java
```diff
@@ -428,7 +427,7 @@ public String[] supportedKeys() {
     }
 
     protected ByteArrayDataOutputStream getOutputStream(Address mbr, int size) {
-        return output_streams.computeIfAbsent(mbr, addr -> new ByteArrayDataOutputStream(size));
+        return output_streams.computeIfAbsent(mbr, __ -> new ByteArrayDataOutputStream(size));
```
rhusar, line 430: Whoa, fun, is this a new convention? Strictly speaking, this would violate the classical naming convention https://www.oracle.com/java/technologies/javase/codeconventions-namingconventions.html

belaban: Fine by me! same as `ignored` instead of `addr`

pruivo: bad habit of mine when I don't need the variable in lambdas.

rhusar: I usually go for ignored, IIUC it's also picked up by IDEs and don't underline as 'unused'.

PR #733, opened 2023-08-14
Mark byteman dependency as Optional

ryanemerson:
[ISPN-15109](https://issues.redhat.com/browse/ISPN-15109) for more context.


belaban:
Need to check if copy-dependencies copies optional JARs, too

PR #742, opened 2023-10-12
JGRP-2736 Use Instant instead of Date

tristantarrant:
https://issues.redhat.com/browse/JGRP-2736


tristantarrant:
@belaban I've introduced a method that formats strings similarly to what `Date.toString()` does but without triggering the TZ caching. Obviously it doesn't adapt to the local timezone, so it's always UTC:
`Thu Oct 12 13:35:51 UTC 2023`

PR #755, opened 2023-12-14
Fix SecurityActions compilation on latest JDK (local variables refere…

rhusar:
…nced from a lambda expression must be final or effectively final)


rhusar:
This fixed the JDK 22 compilation issues.

PR #761, opened 2024-01-16
Synchronize view change and retransmit task

wburns:
https://issues.redhat.com/browse/JGRP-2749

Note, I am not a fan of this way to fix, but there really needs to be some exclusion between these operations to not have dangling data around. To be honest looking at this class there are a lot of places that are referencing mutable collections that appear to be reachable from multiple threads, so I fear this won't fix other possible causes of CME that could happen.

For example another possible CME is if there is an UP event that removes an entry from the recv_table but we are sending a xmit task that is currently iterating upon it.

@belaban let me know how you want to proceed


belaban:
Which scenario are you looking at? Receiving+adding a message in `ReceiverEntry.received_msgs`, and removing it at the same time? This should not be a problem, as `Table` is synchronized...

belaban:
Fixed this by changing MessageCache

PR #766, opened 2024-01-26
[4.x] Fix MessageCacheTest compilation for JGRP-2747.

rhusar:
Fixes regression caused by backport - https://github.com/belaban/JGroups/commit/5d7290d20861936c8a2f939d0870acdbb22f6a35

The `ObjectMessage` was introduced in 5.0 :-)


belaban:
Thanks for catching this!

PR #768, opened 2024-01-26
JGRP-2739 Cache addresses on first access

pruivo:
`CACHED_ADDRESSES` is been overwritten all the time the method is invoked when `filter!=null`. This PR fixes it (assuming it wasn't intentional)


belaban:
Thanks, merged manually!

PR #779, opened 2024-04-05
Throw UnsupportedOperationException for unsupported ForkChannel.connect() methods

cfredri4:
`ForkChannel.connect(String cluster_name, boolean useFlushIfPresent)` and `ForkChannel.connect(String cluster_name, Address target, long timeout, boolean useFlushIfPresent)` are not overridden in ForkChannel and will call subclass JChannel. This change instead makes them throw UnsupportedOperationException.
This change also corrects the javadoc that a closed ForkChannel can be reused (it cannot).


cfredri4:
No longer needed as flush has been removed.

belaban:
Got it

PR #784, opened 2024-04-12
JGRP-2782 - TP: new global config VERSION_CHECK_MICRO should consider Jgroups as incompatible if Micro Version differs

kicktipp:
https://issues.redhat.com/browse/JGRP-2782

Some whitespace was changed too, sorry. Didn't know how to stop my IDE form doing this.

### DIFF
--- a/src/org/jgroups/Global.java
+++ b/src/org/jgroups/Global.java
@@ -4,9 +4,9 @@ import org.jgroups.util.UUID;
 
 /**
  * Global is a JGroups internal class defining global variables.
- * 
+ *
  * @since 2.0
- * @author Bela Ban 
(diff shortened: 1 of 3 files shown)


belaban:
Thanks Vygen! Applied and committed; welcome as a new JGroups dev! :-)

rhusar:
Perhaps for next time: the commit message is useless, SCM best practices tell us to keep formatting changes separate from functional changes, the whitespace around e.g. assignment are inconsistent, some if conditions use block some don't :-D

belaban:
Also, some acronyms <irony>:
* {{pp}}: pretty print
* {{ns}}: nothing special
</irony>

PR #792, opened 2024-04-30
Add explicit module info

bowbahdoe:
Adds an explicit module-info to the final artifact

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -148,9 +148,20 @@
 
     <target name="compile" depends="prepare" description="Compiles all Java files">
         <javac destdir="${compile.dir}" release="11"
-               classpathref="jg.classpath" includeantruntime="false" debug="on"
+               classpathref="jg.classpath"
+               modulepathref="jg.classpath"
+               includeantruntime="false"
+               debug="on"
                includes="org/jgroups/**">
             <src path="${src.dir}"/>
+        </javac>
+
+        <javac destdir="${compile.dir}" release="11"
+               classpathref="jg.classpath"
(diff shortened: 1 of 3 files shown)


belaban:
Hmm... compilation with ant and mavcen succeeds, but IntelliJ complains:
java: module not found: org.apache.logging.log4j.core
java: module not found: org.apache.logging.log4j
java: module not found: org.slf4j

bowbahdoe:
This is perhaps not helpful, but IntelliJ is working fine on my machine. Maybe you need to upgrade your IJ?

belaban:
Still the same problem, I'm on 2023.2.6

bowbahdoe:
Thats the same version as me. 

I nuked my local and re-cloned from github and all was fine. Maybe its an issue as to whether its a "maven project" in intellij's eyes? I'm also on a mac if that ends up mattering.

(I definitely have incentive to figure out whats going on - partially so you don't revert and partially so I can be aware of this in the future)

tristantarrant:
I think that this pr must have a few tests executed with the modular path,
otherwise it's hard to validate its correctness.

Tristan

belaban:
This is driving me crazy and now I have a new error: Exception in thread "main" java.lang.NoClassDefFoundError: production/JGroups/module-info is not a class because access_flag ACC_MODULE is set.

Same issue regardless whether I import the project as maven or from srcs (ant).

I'm removing `module-info.java`, as it prevents me from working. I'll leave the changes in `build.xml` and `pom.xml`.

PR #806, opened 2024-06-18
Fix FD_SOCK2 startup log to show *:port instead of *.port

tristantarrant:
For some reason this always bothers me :)


rhusar (approved):
+1!

belaban:
For some reason I let this one slip... applied now and backported to 5.3.x

PR #810, opened 2024-06-20
JGRP-2809 Allow setting TLS SSLContext programmatically

cfredri4:
This change allows setting TLS SSLContext directly when wanting to configure the SSLContext programmatically


belaban:
OK, but it won't make your code not compile, will it?

rhusar:
> OK, but it won't make your code not compile, will it?

All good here.

cfredri4:
In my mind it comes down to different levels of configuration; the highest level is configuring they keystore/trust store (which was already possible), the lowest level is providing your own socket factory (also possible), and there's the middle ground of providing your own context (which is what this PR enables).

PR #817, opened 2024-07-11
JGRP-2813 GossipRouter Support Certificate Reloading

ryanemerson:
https://issues.redhat.com/browse/JGRP-2813

@tristantarrant This is effectively a port of [ISPN-15916](https://issues.redhat.com/browse/ISPN-15916), can you take a quick look?


belaban:
Applied and backported to 5.3.9

PR #836, opened 2024-09-13
JGRP-2835 Do not log warn during normal shutdown.

pferraro:
and fix interrupt handling.

https://issues.redhat.com/browse/JGRP-2835


belaban:
Applied manually

PR #844, opened 2024-10-11
Was causing undefined class error 

snyderra:
Follow pattern for start and only unregister if jmx is true


belaban:
Thanks for the PR, Rob!

PR #847, opened 2024-10-17
More programmatic setters in TP

cfredri4:


src/org/jgroups/protocols/TP.java
```diff
@@ -207,6 +207,9 @@ public String getBundlerClass() {
     public long             getTimeServiceInterval() {return time_service_interval;}
     public <T extends TP> T setTimeServiceInterval(long t) {this.time_service_interval=t; return (T)this;}
 
+    public boolean          useVThreads()          {return use_vthreads;}
```
rhusar, line 210: That's pretty bad naming. Methods returning a boolean should have names that start with "is" or "has".

src/org/jgroups/protocols/TP.java
```diff
@@ -443,7 +446,10 @@ protected TP() {
     }
 
     public MsgStats                getMessageStats()     {return msg_stats;}
-    public MessageProcessingPolicy msgProcessingPolicy() {return msg_processing_policy;}
+
+    public MessageProcessingPolicy msgProcessingPolicy()                          {return msg_processing_policy;}
```
rhusar, line 450: Very inconsistent naming. The method right above uses proper 'getMessageStats' there's no reason to shorten the name here.

rhusar (changes requested):

src/org/jgroups/protocols/TP.java
```diff
@@ -207,6 +207,9 @@ public String getBundlerClass() {
     public long             getTimeServiceInterval() {return time_service_interval;}
     public <T extends TP> T setTimeServiceInterval(long t) {this.time_service_interval=t; return (T)this;}
 
+    public boolean          useVThreads()          {return use_vthreads;}
+    public <T extends TP> T useVThreads(boolean b) {use_vthreads=b; return (T)this;}
```
rhusar, line 211: Also, "VThreads"? That's not a thing to use in public API.

rhusar (changes requested):

src/org/jgroups/protocols/TP.java
cfredri4: I tried to stay consistent with the version from <=5.3 which useVirtualThreads(), or other existing getters i.e. receiveOnAllInterfaces().
Should I still prefix with "is" for this?

rhusar: The naming in this class is already very inconsistent, e.g. isTrace is both getter and setter but by convention these should be setTrace/isTrace, so looking at past code for inspiration is not great.

Replying on parametric polymorphism for getter/setter is an antipattern.

rhusar: Also, the method name doesn't have to reflect the internal variable this sets (especially since the variable naming conventions in this particular projects are not Java-based..)

src/org/jgroups/protocols/TP.java
cfredri4: It was `use_virtual_threads` in <=5.3 [but was renamed to vthreads in 5.4](https://issues.redhat.com/browse/JGRP-2842?focusedId=25806039&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-25806039)?

src/org/jgroups/protocols/TP.java
cfredri4: Same as comment above (but "get" instead of "is").

cfredri4: Existing code, I only formatted it.

cfredri4: Updated according to comment.

src/org/jgroups/protocols/TP.java
cfredri4: Updated according to comment.

rhusar:
@cfredri4 Should be squashed in a single commit.

PR #853, opened 2024-10-23
Add Javadoc for @org.jgroups.annotations.Property configuration and i…

rhusar:
…ts idiosyncrasies.


rhusar:
Pushed again with updated contract definition of the 'writable'. Please check that.

PR #858, opened 2024-11-07
JGRP-2847 Simplify for custom transports to send to multiple destinations

cfredri4:
This change gives a convenient method to override to allow a custom transport to convert data (to e.g. a direct buffer) only once before sending to multiple destinations


belaban:
Applied manually

PR #859, opened 2024-11-07
JGRP-2848 Composite message additions

cfredri4:
This change borrows the optimization from BatchMessage to not serialize dest and src for each message. And it adds the possibility to add a Collection of messages.


belaban:
Applied manually (instead of rebasing)

PR #860, opened 2024-11-07
JGRP-2849 Allow setting bundler instance before init

cfredri4:
This change allows setting the bundler instance on TP before it is initialized/channel is created


belaban:
Applied

PR #861, opened 2024-11-07
Make MessageFactory static

cfredri4:
This change makes MessageFactory into a static class similar to ClassConfigurator. This greatly simplifies adding custom message types, and also allows custom types to be used in Batch/CompositeMessage which otherwise only allows the default types. This comes with the side effect that it's no longer possible to change how the default message types are generated, but this must be a much less common use case than registering custom types. And as a workaround, a new custom type can be registered instead of altering the default types.


belaban:
Makes a lot of sense; great simplification, too. Thanks!

PR #863, opened 2024-11-08
JGRP-2851 Use thread factory for virtual threads

cfredri4:
This change makes ThreadPool use the supplied ThreadFactory also for virtual threads, which means they will have correct names etc

### DIFF
--- a/src/org/jgroups/util/ThreadCreator.java
+++ b/src/org/jgroups/util/ThreadCreator.java
@@ -31,31 +31,29 @@ public class ThreadCreator {
     }
 
     public static Thread createThread(Runnable r, String name, boolean daemon, boolean virtual) {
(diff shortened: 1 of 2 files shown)


cfredri4:
JDK 17 is LTS, but virtual threads are only available in Loom early access builds for JDK 15/16/17/18 (and as preview feature from JDK 19).
The code previously supported Loom early access builds so I tried to replicate that. But maybe support for Loom early access builds can be dropped and instead virtual threads only need to be supported for real builds (>=JDK19)? This would make DefaultThreadFactory+ThreadCreator simpler.

belaban:
I agree that virtual thread support doesn't need to be in JDKs in which they were preview features. Nobody would run with preview enabled in production...

belaban:
Applied

PR #864, opened 2024-11-14
Minor logging tweaks

cfredri4:
This change makes UnicastHeader/3 toString() consistent with most other headers by wrapping in "[]" And changes two logs for new UNICAST4/NAKACK4 to debug to not be logged by default


belaban:
I don't like the `[` and `]` braces around the headers, as a matter of fact I removed them from `NakackHeader`. If we want braces, this should be done uniformly in `Headers.printHeaders`

PR #865, opened 2024-11-14
JGRP-2848 Composite message additions

cfredri4:
This change borrows the optimization from BatchMessage to not serialize dest and src for each message
This also fixes incorrect size being reported for BatchMessage and in various places where Message.writeToNoAddrs is used.


belaban:
Applied manually

PR #876, opened 2025-01-20
Log dropped sends in Make ReliableMulticast/Unicast

cfredri4:
This change makes it so warnings are logged when sends are dropped by ReliableMulticast/Unicast because send buffers are full.


cfredri4:
Or maybe it should even be logged on error? Because these messages are permanently lost.

PR #878, opened 2025-02-10
JGRP-2864 Avoiding synchronized in BaseServer as it would pin the threads when using virtual threads

ahus1:
This avoids the synchronized keyword

https://issues.redhat.com/browse/JGRP-2864


belaban:
Applied with changes (`synchronized` -> `lock.lock()`)

PR #879, opened 2025-02-10
JGRP-2863 Avoiding synchronized in GMS.installView as it would pin the threads when using virtual threads

ahus1:
JGRP-2863

https://issues.redhat.com/browse/JGRP-2863


belaban:
Appled PR manually, plus made some more changes

PR #887, opened 2025-03-21
JGRP-2868 Util#loadClass throw NPE when getContextClassLoader is null

ryanemerson:
https://issues.redhat.com/browse/JGRP-2868

@belaban We need this for 5.4.x and 5.3.x


belaban:
LGTM, thx

PR #890, opened 2025-03-26
JGRP-2869 Failure detection protocol using existing sockets in TCP/TCP_NIO2

cfredri4:
Here's a proposal for a very simple failure detection protocol which is based on disconnection events from the existing sockets in TCP/TCP_NIO2.
It will only work with TCP/TCP_NIO2 but unlike FD_SOCK2 it does not require any additional ports to be opened.
Please provide feedback and if this makes sense to merge, also let me know how I add any neccessary documentation.


cfredri4:
Closed by d1e781f

PR #892, opened 2025-03-31
Fix send_atomically lock in ReliableUni/Multicast

cfredri4:
The lock needs be taken before incrementing seqno

### DIFF
--- a/src/org/jgroups/protocols/ReliableMulticast.java
+++ b/src/org/jgroups/protocols/ReliableMulticast.java
@@ -696,10 +696,6 @@ public abstract class ReliableMulticast extends Protocol implements DiagnosticsH
     }
 
     protected boolean send(Message msg, Buffer<Message> win, boolean dont_loopback_set) {
-        long msg_id=seqno.incrementAndGet();
-        if(is_trace)
-            log.trace("%s --> [all]: #%d", local_addr, msg_id);
-        msg.putHeader(this.id, NakAckHeader.createMessageHeader(msg_id));
         final Lock lock=send_atomically? win.lock() : null;
         if(lock != null) {
             // As described in doc/design/NAKACK4 ("misc"): if we hold the lock while (1) getting the seqno for a message,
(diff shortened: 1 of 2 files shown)


belaban:
Makes sense - I wonder why I implemented this against my own design / comments... :-) I'll test performance briefly before applying your PR. Thanks!

belaban:
I tested perf (4 node cluster, locally) with tcp-new.xml. `UNICAST4` -> `UPerf`, `NAKACK4` -> `MPerf`:
* +vthreads: `use_vthreads=true`
* -vthreads: `use_vthreads=false`
* +sa: `send_atomically=true`
* -sa: `send_atomically=false`
* old: previosu master
* new: after this PR

### UNICAST4 (UPerf)

| send_atomically | -vthreads | +vthreads |
| ---------------- | ----------| -----------|
|      old -sa           |   112K      |       200K    |
|      old +sa          |   106K      |        196K   |
|      new -sa          |   112K      |       202K    |
|      new +sa         |   107K      |        198K   |



### NAKACK4 (MPerf)

| send_atomically | -vthreads | +vthreads |
| ---------------- | ----------| -----------|
|      old -sa           |   2M         |       2.4M    |
|      old +sa          |    2.06M   |       2.41M   |
|      new -sa          |   2M        |       2.41M    |
|      new +sa         |   2.17M    |        2.47M   |

* Virtual threads speed up performance altogether, regardless of whether `send_atomically` is used, or not
* Enabling `send_atomically` yields slighter lower performance with unicasts (UPerf), but slightly better performance with multicasts (MPerf)
* When `send_atomically` is enabled, both unicasts and multicasts perform slightly better in *new* compared to *old*

cfredri4:
Interesting! My interpretation is that `send_atomically` offers more advantages than disadvantages. Should it be default on instead?

belaban:
For unicasts, it doesn't, only for multicasts. So we could enable `send_atomically` for `NAKACK4` only.

Before doing that, we'd need to do the following (for both unicasts and multicasts):
* Test on AWS and/or GCP
* Run your own perftests

If we get similar results, I can switch the defaults. I don't want to rely on perftesting on my local box only!

PR #893, opened 2025-03-31
Setting for sending views as multiple unicasts

cfredri4:
This change adds a setting to GMS to send new views as multiple unicast messages. This is for use with NAKACK4 to prevent view update messages from being flow controlled to all members.


cfredri4:
Made moot by 0c16aa4

PR #895, opened 2025-04-09
Remove log of dropped sends in Make ReliableMulticast/Unicast

cfredri4:
Reverts #876 since sends can no longer be dropped due to 0c16aa4


cfredri4:
Oops, accidentally opened a new PR. To be continued in #896.

PR #898, opened 2025-04-15
Adds getSocketAddress to IpAddress

cfredri4:
Converting an IpAddress to InetSocketAddress is done in many places. This change adds a getSocketAddress to IpAddress to make this simpler and more efficient.


belaban:
I like the idea of adding method `getSocketAddress()` which will remove some code. I don't like the idea of adding a field `sock_addr` (which btw is assigned but never used!). I don't like carrying the same information in 2 (3) different fields.
I'll apply this PR without adding field `sock_addr`. If you object, let me know why you think `sock_addr` is needed.

cfredri4:
> which btw is assigned but never used
Oops, the intention was to return this for `getSocketAddress()` so as to not needing to create a new `InetSocketAddress` every time. But I see your point, I removed the `sock_addr` now.

cfredri4:
> which btw is assigned but never used

Oops, the intention was to return this for getSocketAddress() so as to not needing to create a new InetSocketAddress every time. But I see your point, I removed the sock_addr now.

belaban:
I merged this PR manually, will close it now.

belaban:
One last thing to do is to make sure that none of the InetSocketAddress constructors do an address lookup (DNS lookup might hang), but I don't remember seeing this.

belaban:
All good!

PR #899, opened 2025-04-23
Use PhysicalAddress in connections and servers

cfredri4:
This change updates connections and servers to instead use the more restrictive PhysicalAddress.

### DIFF
--- a/src/org/jgroups/blocks/cs/BaseServer.java
+++ b/src/org/jgroups/blocks/cs/BaseServer.java
@@ -1,7 +1,7 @@
 package org.jgroups.blocks.cs;
 
-import org.jgroups.Address;
 import org.jgroups.Global;
+import org.jgroups.PhysicalAddress;
 import org.jgroups.annotations.GuardedBy;
 import org.jgroups.annotations.MBean;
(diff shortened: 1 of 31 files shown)


cfredri4:
In almost all cases it was just a trivial change from `Address` to `PhysicalAddress`.
Except for [TUNNEL#L260](https://github.com/belaban/JGroups/pull/899/files#diff-93203ccb199c234ee4ea540b40d66e0807d7d8b3b74836f83d8d430a00f07cdfL260), and I don't fully understand if this is correct.

cfredri4:
The thought was that connections and servers _has_ to use `PhysicalAddress` and so changing to `PhysicalAddress` makes this clearer.
And a related question, what do you think about extending `PhysicalAddress` with `getIpAddress`/`getPort`/`getSocketAddress`? Then connections or servers would not have to downcast anything.
Otherwise feel free to close this. 🙂

belaban:
I've had plans to replace `IpAddress` with `InetSocketAddress`, which I'll take another look at soon. In the same refactoring, I'll see if I need to add more methods to `PhysicalAddress`: [1].
[1] https://issues.redhat.com/browse/JGRP-2876

PR #900, opened 2025-04-25
Changed SyncCounter to allow declared exceptions

mzazaipsc:
https://github.com/jgroups-extras/jgroups-raft/issues/353


belaban:
Hi @mzazaipsc as we discussed, this sort of c hange is *not* suited for a micro release, but only a minor or even a major release.
Also, I haven't seen a final decision on whether or not to use more specific checked exceptions in the signatures rather than `Exception`...

PR #901, opened 2025-05-01
JDBC_PING2 null address with ExtendedUUID

rich-coe:
Upgrading from 5.2.29 to 5.4.6, successfully implemented NAKACK4 and UNICAST4.
Changing from JDBC_PING to JDBC_PING2 get the error "Cannot insert the value NULL into column 'ADDRESS'"
at org.jgroups.protocols.JDBC_PING2.insert(JDBC_PING2.java:353) ~[jgroups-5.4.6.Final.jar:5.4.6.Final]

Investigation showed that Util.addressToString() returning null because addr.getClass() is ExtendedUUID and not a UUID.

### DIFF
--- a/src/org/jgroups/util/Util.java
+++ b/src/org/jgroups/util/Util.java
@@ -1920,7 +1920,7 @@ public class Util {
             return String.format("%s%s:%s:%s", SITE_UUID_PREFIX, su.toStringLong(), su.getName(), su.getSite());
         }
         Class<? extends Address> cl=addr.getClass();
-        if(UUID.class.equals(cl))
+        if(UUID.class.equals(cl) || ExtendedUUID.class.equals(cl))
             return String.format("%s%s", UUID_PREFIX, ((UUID)addr).toStringLong());
         if(IpAddress.class.equals(cl))
             return String.format("%s%s", IP_PREFIX, addr);


belaban:
LGTM. I need to think a bit more about whether the contents of `ExtendedUUID` also need to be stored. That would be terrible, as it breaks backwards-compatibility.
Did you configure JGroups to use `ExtendedUUID`, or is this done by Keycloak by default?

belaban:
I replaced the comparison (`equals()`) with a `Class.isAssignableFrom()`, so other subclasses (such as `FlagsUUID`) are also considered.

belaban:
OK, so it was you who added `ExtendedUUID`, _not_ Keycloak. In any case, this is legal and so it is good you found the bug and fixed it!

belaban:
Hi @rich-coe 
note that [1] was fixed in 5.4.7, which I just released. This is an edge case and highly unlikely to ever occur, but anyway if you get a chance, upgrade to 5.4.7 if you use NAKACK4.

[1] https://issues.redhat.com/browse/JGRP-2875

rich-coe:
Thanks @belaban
I am seeing an issue with NAKACK4 and UNICAST4 I have not tracked down yet.   
I'm getting out-of-memory errror.  There a bunch of threads blocked trying to add data to FixedBuffer blocks.
There are 31 threads in VERIFY_SUSPECT.
I will try 5.4.7 before trying to diagnose more.

PR #910, opened 2025-06-25
Fix invalid ManagedAttribute

cfredri4:
TCP_NIO2 has a field marked as ManagedAttribute which should be ManagedOperation


belaban:
No: this *is* a managed attribute We can therefore do the following:
* `probe TCP_NIO2.reader_idle_time`
* `probe TCP_NIO2.reader_idle_time=100` ## sets to 100ms

cfredri4:
But it returns `void` so it cannot be read?

belaban:
That's the setter only; there's a getter, too

cfredri4:
Oh, ok. But this is the only place out of 500+ places in JGroups that `@ManagedAttribute` is on a void/setter method, that makes it a bit confusing. 🙃

PR #937, opened 2025-09-09
Unchecked exceptions from JChannel

cfredri4:
Following the discussion https://github.com/jgroups-extras/jgroups-raft/issues/353#issuecomment-2838630118, here is a rough initial proposal for not throwing checked exceptions from JChannel.
The idea is that `JGroupsException` is either a wrapper around any checked exception, or the base exception for any specific exception (e.g. `StateTransferException`).
Please feedback or close. 😅

### DIFF
--- a/src/org/jgroups/JChannel.java
+++ b/src/org/jgroups/JChannel.java
@@ -91,15 +91,16 @@ public class JChannel implements Closeable {
         if(create_protocol_stack) {
             try {
                 init(ConfiguratorFactory.getStackConfigurator(Global.DEFAULT_PROTOCOL_STACK));
-            }
-            catch(Exception e) {
-                throw new RuntimeException(e);
+            } catch (RuntimeException e) {
+                throw e;
+            } catch (Exception e) {
+                throw new JGroupsException(e);
             }
         }
     }
 
     /** Creates a {@code JChannel} with the default stack */
-    public JChannel() throws Exception {
+    public JChannel() {
         this(Global.DEFAULT_PROTOCOL_STACK);
     }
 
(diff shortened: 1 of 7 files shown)


cfredri4:
> * Why were SuspectedException and UnreachableException moved from org.jgroups to org.jgroups.blocks?

They were only used by `UnicastRequest` and `MessageDispatcher` also in `org.jgroups.blocks` so I reasoned they belonged there. Right? Wrong?


> * Some exceptions extend RuntimeException, only StateTransferException extends JGroupsException. What's the rationale for this?
> * Also: why not make all exceptions extend RuntimeException? Then we could remove JGroupsException: I don't see why this class was introduced

The idea with `JGroupsException` was that it would be the base exception for any "public" exception (e.g. like [Jackson](https://github.com/FasterXML/jackson-core/blob/3.x/src/main/java/tools/jackson/core/JacksonException.java) or [Kafka](https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/KafkaException.java) or many others does). And `StateTransferException` was a public exception (thrown by `JChannel`) and would thus extend `JGroupsException`.

It can certainly be removed, but then a follow-up question; what should `JChannel` wrap the checked exceptions with in the constructors and `connect`?

> * If we're going to do this, then we should apply this to other classes as well, e.g. RpcDispatcher etc

`RpcDispatcher`, and I guess also `MessageDispatcher`, any others? And I can take a look.

jabolina:
@cfredri4, to have more context, were these changes something that was needed urgently in some manner? If this were needed for a use case, maybe we could find a middle ground? Of course, that means only some methods could change to maintain compatibility, which isn't ideal.

I approached this as "smoothing" the public API, and to me, it makes sense to go all in. But if that breaks the code somehow, it is better to postpone until the next major, which also gives more flexibility if we need to move things around. If this were fixing something broken, I would understand the urge to have it integrated ASAP. On the JGroups Raft side, I believe we can change the interface internally.

cfredri4:
No urgency, just want a smother API to work against. 😉

belaban:
Created https://issues.redhat.com/browse/JGRP-2940 to keep track of this

cfredri4:
Thanks
(but the jira is called _Remove **unchecked** exceptions_)

belaban:
Changed

PR #944, opened 2025-09-30
Consistent tcp/nio cookie warning messages

cfredri4:
Problem: I'm using TCP_NIO2 and seeing sporadic cookie warning messages but cannot identify where the connections are from since peer information is not logged.

This change tweaks the logging of cookie warning message in NioConnection to contain peer information and be consistent with TcpConnection.


cfredri4:
> Why did you remove logging the local address? I'll add this back in

Local address is already added when exception is logged in [NioConnection](https://github.com/belaban/JGroups/blob/master/src/org/jgroups/blocks/cs/NioConnection.java#L202) and [TcpServer](https://github.com/belaban/JGroups/blob/master/src/org/jgroups/blocks/cs/TcpServer.java#L109).

PR #945, opened 2025-10-01
Make ensureCapacity public on ByteArrayDataOutputStream

wburns:
Making the method public so people implementing a custom message marshaller can use this method to ensure the byte[] is large enough before doing other operations.


wburns:
> I suspect this needs to be backported to the `5.4.x` btanch? Just let me know and I'll do it

I don't know yet honestly. I am not sure what version of JGroups we will land on for 16.0, but I figured best to get this in before your next build at least :)

PR #1008, opened 2026-03-24
[JGRP-2987] Interlock in multiple jvm connection open/accept could lead to singleton cluster member even in normal network conditions

elguardian:
Issue: https://redhat.atlassian.net/browse/JGRP-2987

This aims to provide to provide a procedure to avoid rejecting outgoing connections when 2 nodes try to connect to each other
A->B and B->A
When this happen while the algorithm resolves correctly and it is consistent eventually could cause nodes to not detect during bootstrap the cluster and create singleton members.
This provides 2 different locks
1. to hold the critial section of the connection
2. to hold the critical section of producing a connection.

So when and outoing connection and incoming connection happens:
1. the incoming connection comes first. We take that one and the outgoing connection is retrieved by conns.
2. the outgoing connection comes first. In this case if both ends reaches at the same time we need to wait till the connection is established. and then perform the accept /rejection algorithm. The way to detect the deadlock is that the connection will wait for the ack while the lock is still held by the outgoing connection. in that momment we can ensure is a deadlock. Once the deadlock is resolved we can lock the connection production and set as valid. the outgoing connection will wait till the connection is ready.

### DIFF
--- a/src/org/jgroups/blocks/cs/BaseServer.java
+++ b/src/org/jgroups/blocks/cs/BaseServer.java
@@ -24,7 +24,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.locks.Lock;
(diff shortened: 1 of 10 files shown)


belaban:
So this didn't cause any bugs?

PR #1009, opened 2026-04-10
JGRP-2997 Util.parseCommaDelimitedHostsInto returns no addresses when…

rhusar:
… called with StackType.Dual

https://redhat.atlassian.net/browse/JGRP-2997

### DIFF
--- a/src/org/jgroups/util/Util.java
+++ b/src/org/jgroups/util/Util.java
@@ -4096,14 +4096,13 @@ public class Util {
             int port=left_index != -1? Integer.parseInt(t.substring(left_index+1, right_index)) : default_port;
             try {
                 InetAddress[] resolvedAddresses=InetAddress.getAllByName(host);
-                for(int i=0; i < resolvedAddresses.length; i++) {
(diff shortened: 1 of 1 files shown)


rhusar:
So, this fixes the dead code and conditions that can never be met (inet == null  is always false).

However this is never called with StackType.Dual as org.jgroups.protocols.TCPPING#stack_type defaults to Dual BUT it gets overriden in TCPPING#init to specific stack.

rhusar:
The context for this is https://github.com/jgroups-extras/jgroups-kubernetes/pull/309 – Implement support for KUBE_PING to discover IPs for pods with multiple IPs/dualstack especially IPv6 matching the transport stack
 – where I would want to do this without explicit configuration just by consulting the bind address of the transport:

    boolean preferIPv6 = transport.getBindAddr() instanceof Inet6Address;

however, I am discovering issues with how Dual stack is being handled here in jgroups.

belaban:
Removed the inner ports for-loop if add is false, applied manually

PR #1011, opened 2026-04-28
JGRP-3001 Make SelectorProvider configurable via SocketFactory.

rhusar:
Use getSelectorProvider() in both NioServer and NioClient instead of Selector.open() to allow custom NIO providers (e.g. io_uring) without changing the JVM-wide default.

Full backwards compatible; uses default method.

https://redhat.atlassian.net/browse/JGRP-3001

### DIFF
--- a/src/org/jgroups/blocks/cs/NioClient.java
+++ b/src/org/jgroups/blocks/cs/NioClient.java
@@ -7,7 +7,6 @@ import org.jgroups.util.*;
 import java.io.IOException;
 import java.net.InetAddress;
 import java.nio.ByteBuffer;
-import java.nio.channels.Selector;
 
 /**
  * Client to access servers based on TCP_NIO2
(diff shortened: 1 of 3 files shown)


pferraro (approved):

rhusar:
Thanks for the review @pferraro!

@belaban When can we get a release out with this fix included? Thx

belaban:
@rhusar why is this urgent? Do you intend to use this from Wildfly (io_uring)?

rhusar:
> @rhusar why is this urgent? Do you intend to use this from Wildfly (io_uring)?

@belaban Because this currently blocks fixing https://github.com/wildfly/wildfly/pull/19698 as I did not realize that FD_SOCK2 uses NIO and as such breaks the current impl so I need to implement the NIO-based methods as well with TLS-wrapping.

belaban:
LGTM, applied

belaban:
I'll release 5.5.5 tomorrow morning

rhusar:
> I'll release 5.5.5 tomorrow morning

Cool, thanks @belaban !

belaban:
Done, should be in mvn central in a few hours; keep your fingers crossed! :-)

PR #1012, opened 2026-05-04
Optimized sending in TCP_NIO2

cfredri4:
Following up on #952, here's an idea to improve the _sending_ in `TCP_NIO2` by sending direct buffers.

Writing to direct buffers is supposedly slower than writing to heap memory, and in my limited tests I did get _less_ performance if I changed the bundler to write to direct memory.

When sending heap memory, the JDK keeps a single thread local direct ByteBuffer and which the heap memory is copied to before sending.
When sending unicast messages, this realistically means that there will be a single copy of a large chunk of heap memory to direct memory which is about as optimized as it can get.
But when sending multicast messages, the copy will be done _for each destination_ and this is what I try to improve here.

The change is simply to keep a single direct buffer in `NioBaseServer` which is used for multicast sending.
I reuse `use_lock_to_send` not lock around the multicast buffer, this should be improved.

This is not unlike what's mentioned in [JGRP-2998](https://redhat.atlassian.net/browse/JGRP-2998) but I can no longer find the possibility to comment there?

### DIFF
--- a/src/org/jgroups/blocks/cs/BaseServer.java
+++ b/src/org/jgroups/blocks/cs/BaseServer.java
@@ -69,6 +69,7 @@ public abstract class BaseServer implements Closeable, ConnectionListener {
     @Deprecated(since="5.4.4",forRemoval=true)
     protected boolean                         use_acks;
     @ManagedAttribute(description="Log a stack trace when a connection is closed")
+    protected boolean                         use_lock_to_send=true; // e.g. a single sender doesn't need to acquire the send_lock
     protected boolean                         log_details=true;
     protected int                             sock_conn_timeout=1000;      // max time in millis to wait for Socket.connect() to return
     protected boolean                         tcp_nodelay=false;
@@ -105,6 +106,8 @@ public abstract class BaseServer implements Closeable, ConnectionListener {
     public BaseServer       usePeerConnections(boolean flag)        {this.use_peer_connections=flag; return this;}
     public static boolean   useAcks()                               {return false;}
     public BaseServer       useAcks(boolean ignored)                {return this;}
+    public boolean          useLockToSend()                         {return use_lock_to_send;}
+    public BaseServer       useLockToSend(boolean u)                {this.use_lock_to_send=u; return this;}
     public boolean          logDetails()                            {return log_details;}
     public BaseServer       logDetails(boolean l)                   {log_details=l; return this;}
     public int              socketConnectionTimeout()               {return sock_conn_timeout;}
@@ -200,6 +203,26 @@ public abstract class BaseServer implements Closeable, ConnectionListener {
             this.receiver.receive(sender, buf);
     }
 
+    public void send(Collection<? extends Address> dests, byte[] data, int offset, int length) throws Exception {
+        for(Address dest: dests) {
+            try {
+                send(dest, data, offset, length);
+            } catch(Throwable t) {
+                log.error(Util.getMessage("FailureSendingToPhysAddr"), local_addr, dest, t);
+            }
+        }
+    }
+
+    public void send(Collection<? extends Address> dests, ByteBuffer data) throws Exception {
+        for(Address dest: dests) {
+            try {
+                send(dest, data.duplicate());
+            } catch(Throwable t) {
+                log.error(Util.getMessage("FailureSendingToPhysAddr"), local_addr, dest, t);
+            }
+        }
+    }
+
     public void receive(Address sender, DataInput in, int len) throws Exception {
         // https://issues.redhat.com/browse/JGRP-2523: check if max_length has been exceeded
         if(max_length > 0 && len > max_length)
@@ -220,7 +243,7 @@ public abstract class BaseServer implements Closeable, ConnectionListener {
             return;
 
         if(dest == null) {
-            sendToAll(data, offset, length);
+            send(conns.keySet(), data, offset, length);
             return;
         }
 
@@ -246,7 +269,7 @@ public abstract class BaseServer implements Closeable, ConnectionListener {
             return;
 
         if(dest == null) {
-            sendToAll(data);
+            send(conns.keySet(), data);
             return;
         }
 
@@ -510,6 +533,7 @@ public abstract class BaseServer implements Closeable, ConnectionListener {
         return s;
     }
 
+    @Deprecated(since="5.5.6",forRemoval=true)
     public void sendToAll(byte[] data, int offset, int length) {
         for(Map.Entry<Address,Connection> entry: conns.entrySet()) {
             Connection conn=entry.getValue();
@@ -524,6 +548,7 @@ public abstract class BaseServer implements Closeable, ConnectionListener {
         }
     }
 
+    @Deprecated(since="5.5.6",forRemoval=true)
     public void sendToAll(ByteBuffer data) {
         for(Map.Entry<Address,Connection> entry: conns.entrySet()) {
             Connection conn=entry.getValue();
--- a/src/org/jgroups/blocks/cs/NioBaseServer.java
+++ b/src/org/jgroups/blocks/cs/NioBaseServer.java
@@ -7,7 +7,9 @@ import org.jgroups.conf.AttributeType;
 import org.jgroups.util.SocketFactory;
 import org.jgroups.util.ThreadFactory;
 
+import java.nio.ByteBuffer;
 import java.nio.channels.*;
+import java.util.Collection;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.concurrent.locks.Lock;
@@ -36,6 +38,8 @@ public abstract class NioBaseServer extends BaseServer {
 
     protected boolean           use_direct_memory=true;
 
+    protected final Lock        mcast_send_lock=new ReentrantLock();
+    protected ByteBuffer        mcast_send_buf=ByteBuffer.allocateDirect(1024);
 
 
     protected NioBaseServer(ThreadFactory f, SocketFactory sf, int recv_buf_size) {
@@ -87,6 +91,58 @@ public abstract class NioBaseServer extends BaseServer {
             }
         }
         return sb.toString();
+    }
+
+
+    @Override
+    public void send(Address dest, ByteBuffer data) throws Exception {
+        try {
+            super.send(dest, data);
+        } catch(ClosedChannelException | CancelledKeyException ignored) {}
+    }
+
+    @Override
+    public void send(Address dest, byte[] data, int offset, int length) throws Exception {
+        try {
+            super.send(dest, data, offset, length);
+        } catch(ClosedChannelException | CancelledKeyException ignored) {}
+    }
+
+    @Override
+    public void send(Collection<? extends Address> dests, byte[] data, int offset, int length) throws Exception {
+        send(dests, ByteBuffer.wrap(data, offset, length));
+    }
+
+    @Override
+    public void send(Collection<? extends Address> dests, ByteBuffer data) throws Exception {
+        if (data.isDirect()) {
+            super.send(dests, data);
+            return;
+        }
+        boolean use_lock = use_lock_to_send;
+        if (use_lock) {
+            mcast_send_lock.lock();
+        }
+        try {
+            copyToMcastSendBuf(data);
+            super.send(dests, mcast_send_buf);
+        } finally {
+            if (use_lock) {
+                mcast_send_lock.unlock();
+            }
+        }
+    }
+
+    private ByteBuffer copyToMcastSendBuf(ByteBuffer buffer) {
+        if (buffer.remaining() > mcast_send_buf.capacity()) {
+            int newCapacity = Math.max(buffer.remaining(), mcast_send_buf.capacity() * 2);
+            mcast_send_buf = ByteBuffer.allocateDirect(newCapacity);
+        } else {
+            mcast_send_buf.clear();
+        }
+        mcast_send_buf.put(buffer);
+        mcast_send_buf.flip();
+        return mcast_send_buf;
     }
 
 
--- a/src/org/jgroups/protocols/TP.java
+++ b/src/org/jgroups/protocols/TP.java
@@ -11,8 +11,6 @@ import org.jgroups.util.*;
 
 import java.io.DataInput;
 import java.io.InterruptedIOException;
-import java.net.SocketException;
-import java.net.SocketTimeoutException;
 import java.util.*;
 import java.util.concurrent.TimeUnit;
 
@@ -664,17 +662,8 @@ public abstract class TP extends TPConfig implements DiagnosticsHandler.ProbeHan
     }
 
     protected void sendUnicasts(List<PhysicalAddress> dests, byte[] data, int offset, int length) throws Exception {
-        for(PhysicalAddress dest: dests) {
-            try {
-                sendUnicast(dest, data, offset, length);
-            }
-            catch(SocketException | SocketTimeoutException sock_ex) {
-                log.trace(Util.getMessage("FailureSendingToPhysAddr"), local_addr, dest, sock_ex);
-            }
-            catch(Throwable t) {
-                log.error(Util.getMessage("FailureSendingToPhysAddr"), local_addr, dest, t);
-            }
-        }
+        for(PhysicalAddress dest: dests)
+            sendUnicast(dest, data, offset, length);
     }
 
     protected void fetchPhysicalAddrs(List<Address> missing) {
--- a/src/org/jgroups/blocks/cs/NioConnection.java
+++ b/src/org/jgroups/blocks/cs/NioConnection.java
@@ -17,6 +17,7 @@ import java.net.Socket;
 import java.net.SocketAddress;
 import java.net.SocketException;
 import java.nio.ByteBuffer;
+import java.nio.channels.CancelledKeyException;
 import java.nio.channels.ClosedChannelException;
 import java.nio.channels.SelectionKey;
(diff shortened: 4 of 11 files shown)


cfredri4:
And regarding [JGRP-2998](https://redhat.atlassian.net/browse/JGRP-2998), which says:

> When receiving a message, a temporary direct ByteBuffer is created and passed to the socket’s receive operation. The contents are then copied to the byte[] array passed as argument.
>
> Goal:
> ...
> * Make the receivers (UDP.PacketReceiver, TcpConnection, NioConnection) use direct ByteBuffers for reception of messages



That is not correct; after #952 the `MessageReader` keeps a direct buffer for reading and then passes it on directly via `ByteBufferInputStream`.

belaban:
Hey @cfredri4 
what's the error you get when trying to post to JGRP-2998? Are yo logged in?

I just finished JGRP-3002, and am playing with direct/heap memory. Theoretically, direct memory should be faster for sending and receiving of messages as there's no need to copy between kernel and user space. However, `RoundTrip` doesn't show an improvement. I'll run this on GCP to get another data point. Can you try on your end?

Server: `jt RoundTrip -tp nio -server -direct-memory true -direct true -tcp-nodelay true`
Client:  `jt RoundTrip -tp nio -host IP -direct-memory true -direct true -tcp-nodelay true`

* `direct-memory`: make RoundTrip use direct or heap memory to send
* `direct`: make NioTransport use direct/heap memory

Perhaps the payload is too small for direct memory to make a difference... I'll send to add a feature to RoundTrip to define the payload.

In your environment, have you run any perftests between `TCP` and `TCP_NIO2`? E.g. `UPerf` or `MPerf`? In my tests, `TCP` is still a bit faster than `TCP_NIO2`...

cfredri4:
> what's the error you get when trying to post to JGRP-2998? Are yo logged in?

Am logged in but don't even see any comment field or button (the jira looks to have been updated?).


> Theoretically, direct memory should be faster for sending and receiving of messages as there's no need to copy between kernel and user space. 

Yes, but writing to direct memory is supposedly slower, so for many small writes (e.g. bundler) it may be faster to write to heap memory and then do one big copy to direct memory. My crude tests seemed to indicate that.


> However, `RoundTrip` doesn't show an improvement. I'll run this on GCP to get another data point. Can you try on your end? 

I need to prioritize some other things first but hopefully I can test this next week. 


> Perhaps the payload is too small for direct memory to make a difference...

In my tests I send a `CompositeMessage` with 200x 200 byte messages so bundler will do 200x 200 byte writes and then send ~40 kB at once. The `CompositeMessage` is so that I can control the batching instead of relying on the bundler.


> In your environment, have you run any perftests between `TCP` and `TCP_NIO2`?

I've not compared them since we have issues with `TCP` and can only use `TCP_NIO2`.

belaban:
The comment field is at the very end. The new JIRA is frankly speaking a PITA... :-(
If you control your own bundling, perhaps you might want to test using `NoBundler` instead of `PerDestinationBundler`...
I'll take a look at your PR next and will benchmark `TCP` against `TCP_NIO2`.

belaban:
Hmm, I added writing to byte arays, and results look a bit different now:
```
Mac:
----

Benchmark                       (index)  Mode  Cnt      Score      Error  Units
MyBenchmark.testWriteToByteArray           0  avgt    5    622.197 ±   286.105  ns/op
MyBenchmark.testWriteToByteArray           1  avgt    5   3504.386 ±   442.589  ns/op
MyBenchmark.testWriteToByteArray           2  avgt    5  19728.565 ±  3420.744  ns/op
MyBenchmark.testWriteToByteArray           3  avgt    5  37792.424 ±  5713.915  ns/op
MyBenchmark.testWriteToByteArray           4  avgt    5  63256.400 ± 43130.667  ns/op

MyBenchmark.testWriteToDirectBuffer        0  avgt    5     83.567 ±     1.368  ns/op
MyBenchmark.testWriteToDirectBuffer        1  avgt    5   7423.083 ±  1800.816  ns/op
MyBenchmark.testWriteToDirectBuffer        2  avgt    5  14930.562 ±  5066.038  ns/op
MyBenchmark.testWriteToDirectBuffer        3  avgt    5  33938.029 ±  6219.036  ns/op
MyBenchmark.testWriteToDirectBuffer        4  avgt    5  63621.539 ± 18561.704  ns/op

MyBenchmark.testWriteToHeapBuffer          0  avgt    5    601.301 ±   236.488  ns/op
MyBenchmark.testWriteToHeapBuffer          1  avgt    5   1429.085 ±    73.194  ns/op
MyBenchmark.testWriteToHeapBuffer          2  avgt    5  17122.895 ± 10147.181  ns/op
MyBenchmark.testWriteToHeapBuffer          3  avgt    5  37339.985 ±  2081.010  ns/op
MyBenchmark.testWriteToHeapBuffer          4  avgt    5  37288.572 ± 12177.467  ns/op



Linux:
------

Benchmark                            (index)  Mode  Cnt       Score      Error  Units
MyBenchmark.testWriteToByteArray           0  avgt    5    1684.702 ±  396.451  ns/op
MyBenchmark.testWriteToByteArray           1  avgt    5    2650.375 ±  942.302  ns/op
MyBenchmark.testWriteToByteArray           2  avgt    5    4982.381 ±  157.480  ns/op
MyBenchmark.testWriteToByteArray           3  avgt    5   31105.706 ±  827.559  ns/op
MyBenchmark.testWriteToByteArray           4  avgt    5  108546.221 ± 1795.417  ns/op

MyBenchmark.testWriteToDirectBuffer        0  avgt    5     510.570 ±  549.079  ns/op
MyBenchmark.testWriteToDirectBuffer        1  avgt    5    2357.297 ±  402.154  ns/op
MyBenchmark.testWriteToDirectBuffer        2  avgt    5    4697.795 ±   83.806  ns/op
MyBenchmark.testWriteToDirectBuffer        3  avgt    5   31778.311 ± 1108.690  ns/op
MyBenchmark.testWriteToDirectBuffer        4  avgt    5  104638.651 ± 3998.401  ns/op

MyBenchmark.testWriteToHeapBuffer          0  avgt    5    2213.290 ±  761.873  ns/op
MyBenchmark.testWriteToHeapBuffer          1  avgt    5    2476.557 ±  302.952  ns/op
MyBenchmark.testWriteToHeapBuffer          2  avgt    5    4998.133 ±  154.664  ns/op
MyBenchmark.testWriteToHeapBuffer          3  avgt    5   31747.868 ± 2071.303  ns/op
MyBenchmark.testWriteToHeapBuffer          4  avgt    5  105209.192 ± 4012.912  ns/op
```

On Linux, writes to direct byte buffers are consistently the fastest, whereas on MacOS results are inconsistent...

belaban:
I also added a simple test for socket writes (`testSocketWrite()`) via `Socket` (`byte` array), `SocketChannel` (using direct and heap-based `ByteBuffers`). Results:
```
Linux:
------

Benchmark                        (buffer_type)  (size)  Mode  Cnt          Score           Error  Units
SocketBenchmark.testSocketWrite     byte_array     100  avgt    5      24937.678 ±      1678.080  ns/op
SocketBenchmark.testSocketWrite     byte_array    1000  avgt    5     227880.784 ±     89493.962  ns/op
SocketBenchmark.testSocketWrite     byte_array   10000  avgt    5    2330478.024 ±    149031.808  ns/op
SocketBenchmark.testSocketWrite     byte_array   40000  avgt    5  125987185.675 ± 343818686.602  ns/op

SocketBenchmark.testSocketWrite        heap_bb     100  avgt    5      51248.809 ±     12646.922  ns/op
SocketBenchmark.testSocketWrite        heap_bb    1000  avgt    5      72217.684 ±      6823.860  ns/op
SocketBenchmark.testSocketWrite        heap_bb   10000  avgt    5     203528.416 ±     34576.958  ns/op
SocketBenchmark.testSocketWrite        heap_bb   40000  avgt    5     767941.516 ±    440588.523  ns/op

SocketBenchmark.testSocketWrite      direct_bb     100  avgt    5      50953.439 ±      8954.411  ns/op
SocketBenchmark.testSocketWrite      direct_bb    1000  avgt    5      66280.637 ±      5375.797  ns/op
SocketBenchmark.testSocketWrite      direct_bb   10000  avgt    5     196364.324 ±     62421.749  ns/op
SocketBenchmark.testSocketWrite      direct_bb   40000  avgt    5     641132.574 ±    230443.112  ns/op


Mac:
----

Benchmark                        (buffer_type)  (size)  Mode  Cnt        Score         Error  Units
SocketBenchmark.testSocketWrite     byte_array     100  avgt    5    33770.319 ±   89124.157  ns/op
SocketBenchmark.testSocketWrite     byte_array    1000  avgt    5   222194.841 ±  194279.696  ns/op
SocketBenchmark.testSocketWrite     byte_array   10000  avgt    5  2005129.269 ±  163642.612  ns/op
SocketBenchmark.testSocketWrite     byte_array   40000  avgt    5  9274043.481 ± 2312604.306  ns/op

SocketBenchmark.testSocketWrite        heap_bb     100  avgt    5   116504.993 ±    1761.798  ns/op
SocketBenchmark.testSocketWrite        heap_bb    1000  avgt    5   119998.388 ±    2761.280  ns/op
SocketBenchmark.testSocketWrite        heap_bb   10000  avgt    5   265085.491 ±   20153.474  ns/op
SocketBenchmark.testSocketWrite        heap_bb   40000  avgt    5   697649.831 ±  311193.190  ns/op

SocketBenchmark.testSocketWrite      direct_bb     100  avgt    5   111351.192 ±    1059.558  ns/op
SocketBenchmark.testSocketWrite      direct_bb    1000  avgt    5   115922.151 ±    6006.310  ns/op
SocketBenchmark.testSocketWrite      direct_bb   10000  avgt    5   263328.619 ±   67297.208  ns/op
SocketBenchmark.testSocketWrite      direct_bb   40000  avgt    5   676767.836 ±  625250.198  ns/op

```

belaban:
Tested writes to buffers and socket writes on GCP (`n2-standard-8` in zone `europe-west6-a`):
```
MyBenchmark.testWriteToByteArray           0  avgt    5    1101.537 ?   54.996  ns/op
MyBenchmark.testWriteToByteArray           1  avgt    5    2501.166 ?   94.564  ns/op
MyBenchmark.testWriteToByteArray           2  avgt    5    8033.109 ?  460.872  ns/op
MyBenchmark.testWriteToByteArray           3  avgt    5   57755.182 ? 1140.547  ns/op
MyBenchmark.testWriteToByteArray           4  avgt    5  121802.485 ? 9499.786  ns/op

MyBenchmark.testWriteToDirectBuffer        0  avgt    5     343.329 ?  344.352  ns/op
MyBenchmark.testWriteToDirectBuffer        1  avgt    5    2474.240 ?   93.178  ns/op
MyBenchmark.testWriteToDirectBuffer        2  avgt    5    8388.529 ?  231.142  ns/op
MyBenchmark.testWriteToDirectBuffer        3  avgt    5   57251.481 ? 1583.860  ns/op
MyBenchmark.testWriteToDirectBuffer        4  avgt    5  103295.358 ? 5519.217  ns/op

MyBenchmark.testWriteToHeapBuffer          0  avgt    5    1457.293 ?  266.692  ns/op
MyBenchmark.testWriteToHeapBuffer          1  avgt    5    3156.505 ?  639.604  ns/op
MyBenchmark.testWriteToHeapBuffer          2  avgt    5    7671.602 ?  428.014  ns/op
MyBenchmark.testWriteToHeapBuffer          3  avgt    5   54567.510 ? 1482.501  ns/op
MyBenchmark.testWriteToHeapBuffer          4  avgt    5  105491.262 ? 4130.665  ns/op
```

socket writes:
```
Benchmark                        (buffer_type)  (size)  Mode  Cnt         Score          Error  Units
SocketBenchmark.testSocketWrite     byte_array     100  avgt    5     51871.220 ?     1019.100  ns/op
SocketBenchmark.testSocketWrite     byte_array    1000  avgt    5    368674.173 ?    58681.101  ns/op
SocketBenchmark.testSocketWrite     byte_array   10000  avgt    5   4131010.465 ?  1013639.753  ns/op
SocketBenchmark.testSocketWrite     byte_array   40000  avgt    5  19715438.273 ? 14256750.127  ns/op

SocketBenchmark.testSocketWrite        heap_bb     100  avgt    5    351835.480 ?   265134.609  ns/op
SocketBenchmark.testSocketWrite        heap_bb    1000  avgt    5    178634.359 ?    45792.783  ns/op
SocketBenchmark.testSocketWrite        heap_bb   10000  avgt    5    677993.629 ?   152913.174  ns/op
SocketBenchmark.testSocketWrite        heap_bb   40000  avgt    5   1906105.984 ?   284879.743  ns/op

SocketBenchmark.testSocketWrite      direct_bb     100  avgt    5    315415.824 ?    42328.420  ns/op
SocketBenchmark.testSocketWrite      direct_bb    1000  avgt    5    165222.939 ?    16624.817  ns/op
SocketBenchmark.testSocketWrite      direct_bb   10000  avgt    5    543761.084 ?    64791.721  ns/op
SocketBenchmark.testSocketWrite      direct_bb   40000  avgt    5   1725825.767 ?   711418.781  ns/op
```

src/org/jgroups/blocks/cs/NioBaseServer.java
```diff
@@ -90,6 +94,58 @@ public String printBuffers() {
...
+            if (use_lock) {
+                mcast_send_lock.unlock();
+            }
+        }
+    }
+
+    private ByteBuffer copyToMcastSendBuf(ByteBuffer buffer) {
+        if (buffer.remaining() > mcast_send_buf.capacity()) {
+            int newCapacity = Math.max(buffer.remaining(), mcast_send_buf.capacity() * 2);
+            mcast_send_buf = ByteBuffer.allocateDirect(newCapacity);
```
rhusar, line 139: mcast_send_buf grows but never shrinks, is that intentional?

src/org/jgroups/blocks/cs/BaseServer.java
```diff
@@ -69,6 +69,7 @@ public abstract class BaseServer implements Closeable, ConnectionListener {
     @Deprecated(since="5.4.4",forRemoval=true)
     protected boolean                         use_acks;
     @ManagedAttribute(description="Log a stack trace when a connection is closed")
+    protected boolean                         use_lock_to_send=true; // e.g. a single sender doesn't need to acquire the send_lock
```
rhusar, line 72: Incorrent ManagedAttribute description

rhusar (changes requested):
Needs fixes.

src/org/jgroups/blocks/cs/BaseServer.java
rhusar: Actually it's just field added at the wrong line

src/org/jgroups/protocols/TP.java
```diff
@@ -664,17 +662,8 @@ protected void sendToAll(byte[] buf, int offset, int length) throws Exception {
...
-            }
-            catch(SocketException | SocketTimeoutException sock_ex) {
-                log.trace(Util.getMessage("FailureSendingToPhysAddr"), local_addr, dest, sock_ex);
-            }
-            catch(Throwable t) {
-                log.error(Util.getMessage("FailureSendingToPhysAddr"), local_addr, dest, t);
-            }
-        }
+        for(PhysicalAddress dest: dests)
+            sendUnicast(dest, data, offset, length);
```
rhusar, line 666: Why? A single failed sendUnicast() would throw an exception and skip remaining destinations for UDP and other non-TCP transports

rhusar (changes requested):

cfredri4:
@rhusar thanks! This PR should be considered a draft/POC for now until verified that it offers any measurable improvements.
There are some additional things to be fixed apart from what you already pointed out; I don't particularly like the single multicast buffer with lock and would want to come up with something else.

rhusar:
Thanks @cfredri4, makes sense.

belaban:
Let's not worry about these details, I'll review the PR as part of the larger investigation to see where direct ByteBuffers make sense (JGRP-2998). I'm currently testing performance of direct memoey for sending / receiving in RoundTrip on GCP.

belaban:
>> But when sending multicast messages, the copy will be done for each destination 

I don't think we use a copy for each destination! 

If we pass a byte array or a heap-based ByteBuffer, we'll fetch a direct ByteBuffer from a `CarrierThreadLocal` `BufferCache`, copy the contents into it and send the direct buffer. This means, we have a cache of direct buffers possibly on each carrier thread. So, depending on which carrier thread we're sending the message, we might use a different direct buffer on every send. But my guess is that - as the ForkJoinPool prefers using existing threads rather than creating new ones - the same thread will be used most of the time. Of course, unless the FJP has all threads active.

One of the changes in [1] will be to configure bundlers to use direct or heap memory for the output buffer (into which messages are serialized). If direct buffers are configured, the above problem will disappear.

Note that your PR addresses only `TCP_NIO2`; mine will work for `TCP`, too.

[1] https://redhat.atlassian.net/browse/JGRP-2998

belaban:
I guess I misinterpreted what you said... yes, the copy will be done for each destination, but a copy should not be expensive. An *allocation* would be expensive... anyway, I'll keep this in mind when implementing JGRP-2998.

belaban:
@cfredri4 I guess we can close this PR, as JGRP-2998 will make it moot. Thoughts?

cfredri4:
Agree

belaban:
OK. I added the suggestion to collect multiple smaller writes in heap memory and only write to direct memory once to JGRP-2998. Thanks,

PR #1013, opened 2026-05-06
- [JGRP-2971] VERIFY_SUSPECT2 can create spurious suspect events

elguardian:
- Using a timer to run the verification task rather than a separate thread pool. This allows us to use virtual threads if available

Cherry-picked form https://github.com/belaban/JGroups/commit/7fc729e88b81b5119192f4defb9974e2495276fe

Issue in downstream: https://redhat.atlassian.net/browse/JBEAP-26757


belaban:
Hi @elguardian do you need a release of 5.2.30 to mvn central, or does Red Hat use an internal release process?

PR #1018, opened 2026-05-14
JGRP-3005 Add missing i18n key ErrorUpdatingJDBCPINGTable (https://re…

rhusar:
…dhat.atlassian.net/browse/JGRP-3005)

Also fixes missing new lines.

https://redhat.atlassian.net/browse/JGRP-3005


rhusar:
Message is looked up here https://github.com/belaban/JGroups/blob/ba5a490694a9cc615e7e32ae2fb6257336654945/src/org/jgroups/protocols/JDBC_PING.java#L178

PR #1019, opened 2026-05-15
JGRP-3006 Fix log calls where exceptions are passed as format args instead of Throwable

rhusar:
For convenience, commit messages of all commits with reasoninig:

>  JGRP-3006 Fix log calls where exceptions are passed as format args instead of Throwable; e.g. throwing IllegalFormatConversionException
> 
>   When an exception is passed as a varargs format argument (3+ args), it uses the (String, Object...) overload instead of (String, Throwable), causing the stack trace to be silently lost.
>   Restructure these calls to use the (String, Throwable) overload so stack traces are properly logged.
> 
>   Also fixes AddressedDNSResolver which used the invalid %a format specifier on a Throwable, causing an IllegalFormatConversionException that was silently swallowed by the logger.
> 
>   Add guards for debug/trace log calls with string concatenation and Throwable as there is no format version of the method that accepts an exception
> 
>   Replace string concatenation with format strings in debug/trace log calls.
>   Eager string concatenation in log.debug/trace(msg) calls forces toString() and StringBuilder allocation regardless of log level.
>   The varargs overload log.debug/trace(format, args...) checks isDebugEnabled/isTraceEnabled before formatting, avoiding unnecessary work when the level is disabled.

https://redhat.atlassian.net/browse/JGRP-3006

### DIFF
--- a/src/org/jgroups/JChannel.java
+++ b/src/org/jgroups/JChannel.java
@@ -523,7 +523,7 @@ public class JChannel implements Closeable {
         if(target == null)
             target=determineCoordinator();
         if(Objects.equals(target, local_addr)) {
-            log.trace(local_addr + ": cannot get state from myself (" + target + "): probably the first member");
+            log.trace("%s: cannot get state from myself (%s): probably the first member", local_addr, target);
             return this;
         }
         state_promise.reset();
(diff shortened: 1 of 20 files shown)


rhusar:
n.b. none of these fixes are on the hot path (i.e. per message) so, phew, but still needs to be fixed as it's wasting heap allocation.

rhusar:
Also, doesn't make any changes to the format itself - only from the ones that are completely broken.

belaban:
Thanks for catching this. I changed some log statements to log the exception *message* instead of the *stack trace* though, in cases where a lot of logging could ensue logging the full stack trace.

belaban:
Applied manually

PR #1023, opened 2026-06-05
JGRP-3011 SO_LINGER cannot be set to 0 in TcpConnection

rhusar:
https://redhat.atlassian.net/browse/JGRP-3011


belaban:
OK, good to have the same semantics in both `TcpConnection`, `NioConnection` and `GossipRouter`... although it escapes me what the diff is between `linger=off` and `linger=on,timeout=0`: in both cases no lingering occurs.

belaban:
OK, no `TIME_WAIT` state...

rhusar:
Exactly, the linger=0 results in `RST` packet sent to peer (instead of FIN in all other cases) so no `TIME_WAIT`.

PR #1024, opened 2026-06-08
[TEST] EnableSuspectEventsDeterministicEventTest

elguardian:
@belaban @rhusar 

This test makes deterministic the connection closed poising by the test create by @rhusar. The only thing it does is to introduce a delay in the rejoining node to make the other nodes reject the connecttion (from their side) when they are trying to reconnect with the node. The dead connection causes the test to fail. The test is exacly the same as @rhusar did bu just addin this little delay.

Have a look and let me know if you think is worth fixing or not.

### DIFF
--- a/conf/scripts/EnableSuspectEventsDeterministicEventTest/testRepeatedLeaveAndJoin.btm
+++ b/conf/scripts/EnableSuspectEventsDeterministicEventTest/testRepeatedLeaveAndJoin.btm
@@ -1,0 +1,59 @@
+
+## Trace + delay: log the critical points and delay the rejoining node's
+## replaceConnection to widen the race window for the concurrent connect.
+
+RULE T1_getConnection
+CLASS BaseServer
+METHOD getConnection
+AT ENTRY
+IF TRUE
+DO
+    System.out.println("--> " + Thread.currentThread().getName().substring(Thread.currentThread().getName().lastIndexOf(",") + 1) + " [T1] getConnection(" + $1 + ") node=" + $0.localAddress() + " thread=" + Thread.currentThread().getName())
+ENDRULE
+
+RULE T4_handleOutgoing
+CLASS BaseServer
+METHOD handleOutgoingConnection
+AT ENTRY
+IF TRUE
+DO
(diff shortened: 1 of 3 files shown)


rhusar:
@elguardian Good stuff, thanks for doing this. I can't believe this is still broken.

Could you please re-test this with changes that were not accepted in https://github.com/belaban/JGroups/pull/978 to see whether that fixes it completely?

elguardian:
@rhusar That won't work (tried). The logic is a problem of the connection pool. If clustering team want to fix this, I can see two options:
1. rework the incoming/outgoing conections to support critical sections correcly.
2. If that is complex work, have a connection reaper or validator.

In this case the system fails because of GMS timeout.... but it is enough to create inesatbilities in the CI. I know I am insistent o this issue... but I consider quite critial and it is causing plenty of problems in CI.

elguardian:
@rhusar this is more trivial fix that trying to fix the logic of incoming outgoing connections. It just remove stale connections from the pool when they are detected.
https://github.com/belaban/JGroups/pull/1026

belaban:
Before we get into changing code, I want to confirm that the code is actually broken and what the exact issue is. 

Your test
* Disconnects A from ABC, so the view is BC for B and C
* A reconnects to BC, but the byteman script prevents this by delaying the connection establishment (the discovery attempt)
* This means A discovers no members and therefore becomes a singleton
* Next you wait for 10s for the views to converge, but this won't happen as (1) A has already successfully connected by becoming a singleton cluster and (2) `MERGE3` takes between 10 and 30s to kick in, so it won't kick in before you check for the view convergence. Note that - if you lowered the merge timeouts, the test would pass. Well, actually, it wouldn't because you make the incorrect assumption that a merge between A and BC results in ABC, but this may not be the case as the default policy orders members according to UUID, so you might have BAC as merged view.

This is as designed. Let me know if you disagree, and if so what the exact error is.

belaban:
I also fail to see how this relates to `enable_suspect_events` being `true`: if you disable this and add `FD_SOCK2`, the behavior is the same

belaban:
Closing this one, as this works as designed (see my explanation above). If you manage to come up with code that reproduces a bug caused by `enable_suspect_event` being `true`, please open a new PR.

PR #1026, opened 2026-06-15
Fix stale connection handling in BaseServer.getConnection() and handleIncomingConnection()

elguardian:
getConnection(): when the locked path finds a connection in the conns map that is not connected (stale), remove and close it before creating a new one.

handleIncomingConnection(): when the "bigger address wins" logic would reject an incoming connection, first check if the existing connection is stale (last_access older than sock_conn_timeout). If stale, accept the incoming and replace the dead connection instead of rejecting. This prevents the infinite rejection loop where a node cannot rejoin the cluster because a peer holds a stale connection entry from a previous epoch.

### DIFF
--- a/src/org/jgroups/blocks/cs/BaseServer.java
+++ b/src/org/jgroups/blocks/cs/BaseServer.java
@@ -302,6 +302,10 @@ public abstract class BaseServer implements Closeable, ConnectionListener {
         try {
             if(connected(conn=conns.get(dest)))
                 return conn;
+            if(conn != null) {
+                conns.remove(dest);
+                Util.close(conn);
+            }
             conn=createConnection(dest);
             handleOutgoingConnection(dest, conn);
         }
@@ -365,8 +369,14 @@ public abstract class BaseServer implements Closeable, ConnectionListener {
         Lock lock=getLock(peer_addr);
         lock.lock();
         try {
-            boolean conn_exists=hasConnection(peer_addr),
+            Connection existing=conns.get(peer_addr);
+            boolean conn_exists=connected(existing),
               replace=conn_exists && local_addr.compareTo(peer_addr) < 0; // bigger conn wins
+
+            if(!replace && conn_exists && isStale(existing)) {
+                replace=true;
+                log.trace("%s: existing connection to %s is stale, replacing", local_addr, peer_addr);
+            }
 
             if(!conn_exists || replace) {
                 notifyConnectionEstablished(conn);
@@ -384,6 +394,14 @@ public abstract class BaseServer implements Closeable, ConnectionListener {
         }
         if(close_conn)
             Util.close(conn); // closing the connection outside the lock scope (might block if TLS)
+    }
+
+    public boolean isStale(Connection conn) {
+        if(conn == null)
+            return false;
+        long now=conn.getTimestamp();
+        long threshold=TimeUnit.NANOSECONDS.convert(sock_conn_timeout, TimeUnit.MILLISECONDS);
+        return now - conn.last_access > threshold;
     }
 
     public BaseServer addConnectionListener(ConnectionListener cl) {
--- a/tests/junit-functional/org/jgroups/tests/StaleConnectionTest.java
+++ b/tests/junit-functional/org/jgroups/tests/StaleConnectionTest.java
@@ -1,0 +1,86 @@
+package org.jgroups.tests;
+
+import org.jgroups.Global;
+import org.jgroups.blocks.cs.Connection;
+import org.jgroups.blocks.cs.TcpServer;
+import org.jgroups.stack.IpAddress;
+import org.jgroups.util.Util;
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.Test;
+
+import java.net.InetAddress;
+
+/**
+ * Tests for stale connection handling in BaseServer.
+ * <ul>
+ *   <li>Test 1: getConnection() cleans up a stale outgoing connection</li>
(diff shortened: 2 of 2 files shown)


belaban:
I'll look at this once we've resolved the conversation in #1024. But a quick glance shows that the fix won't work if connection reaping is disabled, because the timestamp in a connection will not be updated in this case. 

Also, if you have a time service enabled, whose interval is greater than `sock_conn_timeout`, the definition of stable will fail.

elguardian:
@belaban it does not try to solve the problem incoming / outoing conection logic as it was stated before that is a far too risky fix. This tries to get rid of some problems I am seeinng in the CI like this https://redhat.atlassian.net/browse/JGRP-3013 (which is the same problem over and over again- There are other test too)... it is more like a reacting thing. We can discuss how to reap the connection when certains things are not configured but I think this is safer approach.

This works (I did run during the weekend with no failures)... usually you reach a failure without touching the test in 3-4 hours more or less.

This issue is extremely difficutl to reproduce as it requires a connection rejected replacement in the node joining and an outoing conection from the coordinator.. so the joining node tries to reconnect indefenetly and the coordinator keeps rejecting because it has a good connection in its pool and the conflict connection resolution keeps working in favor of the coordiantor as its port win the race... so the joining node will never joing the cluster.

anyway let me know your thoughts.... I have a bunch of data and test regarding this problem.

belaban:
Your second but last paragraph: can we write a reproducer? E.g. we can inject addresses which always makes the joiner's address lower than the coordinator's, so that the coordinator always wins.

If you provide a more concise description of how this can be reproduced, I'll take a look at creating a reproducer. E.g. how can a connection in the client result in rejection etc.

I'm extremely hesitant to change code which works well, only to support a feature which tries to derive a perfect failure detector from connection management (which can be unreliable)...

belaban:
In your opinion, would the following reproduce the issue?
* A has port 1000 (local port to  connect to B is 1500), B 2000 (local port to connect to A is 2500)
* B sends a message to A, therefore has a connection B -> A
* A clears its connection table
* A sends a message to B, connects at 1500; this should be rejected by B

I'm writing a reproducer in ServerTests, trying to see if this fails.

belaban:
I can actually reproduce this but only if I remove a connection on the client side *without* closing it. I don't see how this could happen, as connections are always closed when removed...

elguardian:
@belaban you can check the logs in the jira I did post. the close gracefully during tons of times I think is signaling that. and that is the reason why sometimes the cluster times out no matter what you do. I have actually have a PR that consistently reproduce the issue in the CI.
https://ci.wildfly.org/buildConfiguration/WF_PullRequest_BootableJarLinuxJdk25/565911
https://github.com/wildfly/wildfly/pull/19428
If you retest this almost 95 % of the time fails. The test is simple restart ot the node.

belaban:
I don't have a team city login

elguardian:
@belaban you can log anonymously... the logs are attached here as well https://redhat.atlassian.net/browse/JGRP-3013

belaban:
@elguardian You mentioned a reproducer with a 95% quote. That's what I'm interested in, not logs. The logs don't tell me how this open-close ping pong game started..

src/org/jgroups/blocks/cs/BaseServer.java
```diff
@@ -386,6 +396,14 @@ public void handleIncomingConnection(Address peer_addr, Connection conn) throws
             Util.close(conn); // closing the connection outside the lock scope (might block if TLS)
     }
 
+    public boolean isStale(Connection conn) {
```
rhusar, line 399: If anything wrong encapsulation, this would be a method on a connection, not the BaseServer.

src/org/jgroups/blocks/cs/BaseServer.java
```diff
@@ -386,6 +396,14 @@ public void handleIncomingConnection(Address peer_addr, Connection conn) throws
...
+        if(conn == null)
+            return false;
+        long now=conn.getTimestamp();
+        long threshold=TimeUnit.NANOSECONDS.convert(sock_conn_timeout, TimeUnit.MILLISECONDS);
```
rhusar, line 403: This is conflating terms completely, sock_conn_timeout has completely different meaning and has nothing to do with connection 'freshness'.

rhusar, line 403: This defaults to 1 second – so this would be closing and opening a lot of connections, that makes no sense.

src/org/jgroups/blocks/cs/BaseServer.java
```diff
@@ -386,6 +396,14 @@ public void handleIncomingConnection(Address peer_addr, Connection conn) throws
...
+        return now - conn.last_access > threshold;
```
rhusar, line 404: Direct access to field... please no.

src/org/jgroups/blocks/cs/BaseServer.java
```diff
@@ -386,6 +396,14 @@ public void handleIncomingConnection(Address peer_addr, Connection conn) throws
...
+    }
```
rhusar, line 405: This implementation doesn't work, as org.jgroups.blocks.cs.Connection#updateLastAccessed is never updated by default org.jgroups.blocks.cs.BaseServer#conn_expire_time is zero.

rhusar (changes requested):
The impl is not correct – but in general, the whole idea of stale connection is not clear. Why would we ever do it? Why would we care the connection is open and unused for few seconds?

belaban:
I'm closing this PR as the fix is flawed, as Rado and I explained. If you manage to come up with a reproducer, I'm interested in fixing this.

PR #1029, opened 2026-06-16
Create Ansible setup for JGroups

jabolina:
This one is only a small setup. It installs the dependencies and pushes the local JGroups build remotely to all nodes. I've copied the inventory template we have in IspnPerf, too.

I haven't included any playbook that runs something, and also, no playbook that stops all Java processes in all nodes.
Let me know if you are interested in any of that.


belaban:
Thanks Jose, I'll have a look

PR #1032, opened 2026-06-19
JGRP-3015 Remove dangling connection created during concurrent shutdo…

rhusar:
…wn in BaseServer.send()

https://redhat.atlassian.net/browse/JGRP-3015

### DIFF
--- a/src/org/jgroups/blocks/cs/BaseServer.java
+++ b/src/org/jgroups/blocks/cs/BaseServer.java
@@ -239,6 +239,13 @@ public abstract class BaseServer implements Closeable, ConnectionListener {
             removeConnectionIfPresent(dest, conn);
             throw ex;
         }
+        finally {
+            // Remove connection that may have been created after stop() already cleared conns (JGRP-3015)
+            if(!running()) {
+                log.trace("%s: removing dangling connection to %s created during shutdown", local_addr, dest);
+                removeConnectionIfPresent(dest, conn);
+            }
+        }
     }
 
     public void send(Address dest, ByteBuffer data) throws Exception {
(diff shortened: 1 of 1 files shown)


rhusar:
Why I suspect this is the issue we have been seeing for a long time and made this a blocker: 

The dangling connection keeps the TCP socket ([::1]:7600) alive from old node-1 to node-2. From node-2's perspective, old node-1 still appears connected (the process holding that connection is still running, we called undeploy, not server stop). This also prevents FD_SOCK2 from suspecting old node-1 and node-2's GMS doesn't expel it from the view.  On rejoin, when new node-1 sends TCPPING discovery, node-2 may still have old node-1 in its view as coordinator, causing the  discovery/join to fail  or the stale connection state may prevent the response from reaching new node-1 entirely (the response goes to old node-1's socket, not new node-1's server and its lost and singleton cluster is formed, causing the test to fail). 

With this fix when the discovery response needs to go to [::1]:7600, a fresh connection is created to new node-1's server.

belaban:
Can you describe exactly what you think causes this issue? E.g.

* node-1 calls BaseServer.stop()
* node-1 creates a new connection to node-2
* assertion: node-1 has a connection to node-2 and vice versa
* how does this relate to FD_SOCK2? The server in this protocol should not send spurious data to node-2 and therefore keep the connection closed...
* node-1 creates a new server and calls server.start()

belaban:
OK, the change is uncritical & small, I applied it, thanks,

rhusar:
> Can you describe exactly what you think causes this issue? E.g.
> 
> * node-1 calls BaseServer.stop()
> * node-1 creates a new connection to node-2
> * assertion: node-1 has a connection to node-2 and vice versa
> * how does this relate to FD_SOCK2? The server in this protocol should not send spurious data to node-2 and therefore keep the connection closed...
> * node-1 creates a new server and calls server.start()

@belaban Yes, in other words (from the Jira):

  1. TcpServer.stop() sets running=false (line 89), then closes server socket, then calls super.stop() which closes all connections and clears conns
  2. A concurrent send() thread has already passed validateArgs() and is about to call getConnection(dest)
  3. getConnection(dest) finds conns empty (just cleared), creates a new TcpConnection, connects it, and adds it to conns
  4. send() completes - this connection is never closed

You are right about FD_SOCK2 here, FD_SOCK2 itself doesn't create the dangling connection, the dangling connection is in the transport's TcpServer, created by a concurrent send() with stop().

I haven't check this in detail, but my hunch is that when new node-1 reconnects  on the same port, node-2 sends the discovery response over the stale incoming connection to old node-1's dangling socket instead of creating a fresh connection to new node-1's server. Never received = singleton cluster.

PR #1033, opened 2026-06-25
JGRP-3017 Set SO_REUSEADDR on server sockets before binding

rhusar:
https://redhat.atlassian.net/browse/JGRP-3017


rhusar:
FYI @pferraro

PR #1034, opened 2026-06-25
CI: Add JDK 26 (latest GA) to the build matrix

rhusar:


rhusar:
FWIW I didn't include 27-ea as that seems a bit of an overkill.

belaban:
Yes, we can do that later

PR #1040, opened 2026-07-19
Migrate JGroups unit tests from Ant+TestNG to Maven Surefire+TestNG

akbar1214:
This PR migrates JGroups unit test execution from the Ant-based TestNG runner to Maven Surefire + TestNG, while keeping the existing directory structure and Ant build intact.
What changed
- Added a byteman.version property and centralized both byteman and byteman-bmunit dependencies on it.
- Added 12 TestNG XML suite files under src/test/resources/testng/, mirroring the existing Ant suites:
- functional, encrypt, byteman, jdbc, stack-independent, time-sensitive, udp, udp-new, tcp, tcp-new, tcp-nio, tcp-nio-new
- Each suite defines the correct groups, exclusions, parallelism, thread count, and channel.conf parameter where needed.

- Surefire configuration
  - Replaced the placeholder Surefire config with 12 self-contained executions.
  - Each execution supplies its own suite XML, system properties, port ranges, and JVM arguments.
  - Removed the <systemPropertiesFile>build.properties</systemPropertiesFile> dependency and inlined test-relevant properties.
  - Added tests/util as a test source so the custom JUnitXMLReporter listener is available to Surefire.
  - Suppressed the default default-test execution so mvn test runs only the defined suites.

- Removed maven-antrun-plugin
  - Replaced the Ant-based XML schema generation with exec-maven-plugin.
  - Copied the generated XSD files with maven-resources-plugin.
  - Removed the plugin from <pluginManagement>.
  - Keystore files are created on demand by the encrypt tests via KeyStoreGenerator, so no separate lifecycle step is needed.
 
How to run
  **mvn test**                                                 # all suites sequentially (ant target `all-tests` didn't run jdbc which is bit different currently)
  **mvn surefire:test@<suite-name>**        # single suite, e.g. functional, encrypt
  
  Part of JGRP-2499

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -144,6 +144,27 @@ total ordering and so on.
 JGroups ships with multiple sample configurations, e.g. `tcp.xml`, which uses TCP instead of UDP as transport
 protocol. Consult the manual (link below) for more details.
 
+== Running the tests
+
+The unit tests are executed with Maven:
+
+[source,bash]
+----
+mvn test
+----
+
+This runs all TestNG suites (functional, encrypt, byteman, jdbc, stack-independent,
+time-sensitive, udp, udp-new, tcp, tcp-new, tcp-nio, tcp-nio-new).
+
+To run a single suite:
+
+[source,bash]
+----
+mvn surefire:test@functional
+----
+
+Replace `functional` with any of the suite names listed above.
+
 == Links
 * Web site: link:$$http://jgroups.org$$[http://www.jgroups.org]
 * Manual: link:$$http://www.jgroups.org/ug.html$$[http://www.jgroups.org/ug.html]
--- a/pom.xml
+++ b/pom.xml
@@ -278,6 +278,7 @@
                           <source>tests/byteman</source>
                           <source>tests/junit</source>
                           <source>tests/junit-functional</source>
+                           <source>tests/util</source>
                           <!-- tests/other and tests/perf are included in the normal sources -->
                           <source>tests/stress</source>
                        </sources>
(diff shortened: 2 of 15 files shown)


belaban:
Hi Akbar
thanks for your contribution! I hope to have time to take a look at this this week.

rhusar:
Letting CI run in the meantime - but note this does not actually run the tests.

akbar1214:
@belaban by any chance did you get time to look into this ?

belaban:
Not yet, it's on my todo list...

belaban:
I took a quick look: `mvn surefire:test@functional` has the right number of tests (1701) but 4 failure compared to 0 with ant.
I noticed that `UnicastRequestTest` has the following failure:
```
[ERROR] org.jgroups.blocks.UnicastRequestTest.testMissingResponseDueToMergeViewUnicast -- Time elapsed: 0.136 s <<< FAILURE!
java.util.concurrent.TimeoutException: 
Timeout 10000 kicked in, views are:
A: [A|3] (4) [A, A, B, B]
B: [A|3] (4) [A, A, B, B]
```

It is almost as if the `singleThreaded=true` directive is ignored and multiple tests are run in parallel.
I'll investigate more next week...

belaban:
This surefire output (HTML file) looks hideous! Is there a way to get output from `JUnitXmlReporter` instead?

belaban:
The ant testsuite still works, so I'll integrate your PR. I can fix the errors later, but at least we've integrated your code and can take it from there

PR #1043, opened 2026-07-28
[5.5.x] JGRP-3030 Fix view/membership inconsistencies due to repeated volatile reads.

pferraro:
https://redhat.atlassian.net/browse/JGRP-3030


belaban:
Closed as per explanation in JGRP-3030

PR #1044, opened 2026-08-03
Fix in flight message drop issue happening due to key rotation

akbar1214:
## Summary

In `ASYM_ENCRYPT`, when the shared group key is rotated (member leaves with `change_key_on_leave`, coordinator change, or merge), any message still in flight that was encrypted with the *previous* key is **silently dropped** by the recipient. The sender is never told, and the application never sees the message.

Root cause: `cacheGroupKey()` is only ever invoked *after* `secret_key` has already been overwritten with the new key, so `key_map` ends up holding the (current version → current key) mapping — never the old one. But `Encrypt.decrypt()` only consults `key_map` for a message whose header version differs from the current `sym_version`, i.e. it looks up the **old** key, which is never present. The message is discarded with a `null` return.

## Reproduction

Deterministic regression test: `tests/junit-functional/org/jgroups/protocols/EncryptOldKeyCacheTest.java`

- Install the first group key (version V1); encrypt a payload.
- Rotate to a new group key (version V2) via `createNewKey()`.
- Attempt to decrypt the still-in-flight V1 message.

Fails before fix: the message is dropped (`key_map` holds only V2). 
Passes after fix: the V1 message is recovered with the cached old key.

Run:

```
mvn test-compile surefire:test@encrypt -Dtest=EncryptOldKeyCacheTest
```

## Fix

Cache the previous key in `key_map` **before** overwriting `secret_key`/`sym_version`, in both `createNewKey()` and `installSharedGroupKey()`. The old keys are only used for in-flight messages from current-view members (`inView()` guard is unchanged), so the fix restores delivery of legitimate in-transit messages without allowing non-members to authenticate with stale keys.


belaban:
While the message is dropped by the receipient, it will get retransmitted by `UNICAST3` and - this time - it will get passed up the stack to `UNICAST3` and delivered, as the correct decryption key is used.

Your test doesn't use a JChannel and therefore doesn't take retransmission into account.

I added looking at this PR onto my todo list; your suggested changes might make sense as they make the code more robust, but the unit test does not map to a real-world scenario.

akbar1214:
Your observation is correct. So I have removed the test.

Also on a separate note I have removed the dummy method which is not needed now(I have test in local all 6 classes) and seems like TestNG has fixed this issue.

belaban:
Thanks for the PR! LGTM and all unit tests passed successfully. Merged and pushed to `master`.

belaban:
Backported to the 5.5.x and 5.4.x branches

PR #1053, opened 2026-08-26
Restore Ant-style test reports

akbar1214:
Restore generated test report behaviour


belaban:
Executing `mvn clean surefire:test@functional` doesn't create any `test-results/html` directory...

belaban:
Applied the PR anyway - the ant unit test and report still works

PR #1055, opened 2026-09-02
JGRP-3040 Generate native metadata for graalvm dynamically

tristantarrant:
https://redhat.atlassian.net/browse/JGRP-3040


belaban:
Doesn't work with ant: `ant clean all-tests reports` doesn't find `target/classes`. Added this to `build.xml`.

belaban:
Merged manually

